PR #1, opened 2016-04-01
[WIP] Draft Proposal

kellrott:
This is a task execution proposal based on a loose translation of the Google Pipeline API. Original version found at:
- https://cloud.google.com/genomics/reference/rest/v1alpha2/pipelines
- https://cloud.google.com/genomics/reference/rest/v1alpha2/operations
With example usage found at
https://github.com/googlegenomics/pipelines-api-examples

### DIFF
--- a/proto/task_execution.proto
+++ b/proto/task_execution.proto
@@ -1,0 +1,109 @@
+
+
+syntax = "proto3";
+
+package ga4gh_task_exec;
+
+import "google/api/annotations.proto";
+
+message FileParameter {
+  string path = 1;
+  string storagePath = 2;
+}
+
+message DockerExecutor {
+  string imageName = 1;
+  string cmd = 2;
+}
+
+message Disk {
+  string name = 1;
+  uint32 sizeGb = 2;
+  string source = 3;
+  bool autoDelete = 4;
+  bool readOnly = 5;
+  string mountPoint = 6;  
+}
+
+message Resources {
+  uint32 minimumCpuCores = 1;
+  bool preemptible = 2;
+  uint32 minimumRamGb = 3;
+  repeated Disk disks = 4;
+  repeated string zones = 5;
+}
+
+message Task {
+  string name = 1;
+  string projectId = 2;
+  string description = 3;
+  repeated FileParameter inputParameters = 4;
+  repeated FileParameter outputParameters = 5;
+  Resources resources = 6;
+  string taskId = 7;
+  DockerExecutor docker = 8;  
+}
+
+message TaskRun {
+  repeated FileParameter taskArgs = 1;
+  string taskId = 2;
+  string ephemeralTask = 3;
+}
+
+message TaskId {
+  string value = 1;
+}
+
+message TaskOpId {
+  string value = 1;
+}
+
+message TaskOp {
+  string name = 1;
+  map<string,string> metadata = 2;
+  bool done = 3;
+}
+
+service TaskService {
+  rpc Create(Task) returns (Task) {
+    option (google.api.http) = {
+      post: "/v1/tasks"
+      body: "*"
+    };
+  }
+  
+  rpc Delete(TaskId) returns (TaskId) {
+    option (google.api.http) = {
+      delete: "/v1/tasks/{value}"
+    };    
+  }
+  
+  rpc Get(TaskId) returns (TaskId) {
+    option (google.api.http) = {
+      get: "/v1/tasks/{value}"
+    };    
+  }
+  
+  rpc Run(TaskRun) returns (TaskOpId) {
+    option (google.api.http) = {
+      post: "/v1/tasks:run"
+      body: "*"
+    };    
+  }
+}
+
+
+service TaskOpService {
+  rpc Cancel(TaskOpId) returns (TaskOpId) {
+    option (google.api.http) = {
+      delete: "/v1/taskop/{value}"
+    };
+  }
+  
+  rpc Get(TaskOpId) returns (TaskOpId) {
+    option (google.api.http) = {
+      get: "/v1/taskop/{value}"
+    };
+  }
+  
+}
--- a/proto/task_execution.proto
+++ b/proto/task_execution.proto
@@ -4,25 +4,30 @@ syntax = "proto3";
 
 package ga4gh_task_exec;
 
+// Import HTTP RESTful annotations
 import "google/api/annotations.proto";
 
+/// Mapping long term file storage to path used in docker container
 message FileParameter {
-  string path = 1;
-  string storagePath = 2;
+  string path = 1;        /// Path inside the docker container
+  string storagePath = 2; /// Path of the data's long term storage
 }
 
+/// A command line to be executed and the docker container to run it
 message DockerExecutor {
-  string imageName = 1;
-  string cmd = 2;
+  string imageName = 1;   /// Docker Image name
+  string cmd = 2;         /// The command to be executed
 }
 
+/// Attached disk request. 
 message Disk {
-  string name = 1;
-  uint32 sizeGb = 2;
-  string source = 3;
-  bool autoDelete = 4;
-  bool readOnly = 5;
-  string mountPoint = 6;  
+  string name = 1;        /// Name of attached disk
+  uint32 sizeGb = 2;      /// Minium size
+  string source = 3;      /// Source data, this would refer to an existing disk the execution engine
+                          /// could identify. Leave blank if is to be a newly created disk
+  bool autoDelete = 4;    /// Automatically delete after usage
+  bool readOnly = 5;      /// mount into docker as read only
+  string mountPoint = 6;  /// mount point for disk inside the docker container
 }
 
 message Resources {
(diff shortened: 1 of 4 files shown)


geoffjentry:
@dionloy - wanted to make sure you saw this

geoffjentry:
@rnpandya Wanted to make sure you saw this PR as well

kellrott:
This initial version was based pretty strictly on the Google Pipeline API, as it exists right now. 
But thinking about it, if we added a 'dependsOn' field to the Task message, which would allow for detailing a list of existing task op IDs, you could add the concept of a dependency graph to task execution. To build the job DAG, you'd start by creating the first tasks, getting their taskOp ids, then submitting the subsequent tasks with those ids in their dependsOn fields. 
That would a new requirement for systems that would implement the API. But it could allow for workflow implementation to do less waiting. Right now a CWL or WDL work engine would have to submit single tasks, wait till completion, then submit the dependent jobs. With the 'dependsOn' they could submit the whole chain, and then just wait for the final elements.

dionloy:
Hi Kyle!

What is the end goal / product for this specific API?  If it's task
execution, the current proposal appears to be appropriate. If it is a
complete workflow / pipeline solution, then the 'dependsOn' field only gets
us halfway there while significantly increasing the complexity of
implementation.

CWL and WDL work engines don't need a dependency graph executor, as they
already have them written. As it stands, it's relatively easy to add
execution backends (AWS, Sun Grid Engine, Local, etc ...) to those
controllers since they are not dependent on the task executing backend
supporting a full dependency graph evaluation engine.

On the other hand, if the intent is to replace a CWL or WDL execution
engine entirely (I have no particular opinion on this direction =) ), then
we should go beyond just 'dependsOn' and replace the need for any sort of
CWL / WDL execution engine.  As it stands, with 'dependsOn', you'd still
need CWL or WDL to have their own graph dependency for all other non-GA4GH
task execution backends.

kellrott:
Regarding 'dependsOn', I was curious if having that kind of concept avalible would make it easier to implementing work engines. @tetron ?

One thing that may be more important for work engines,would be to give them ability to stack multiple DockerExecutor statements in a single task. So if the work engine needed to do some sort of file preparation before running the 'user' task code, and then do some sort of post analysis (like parsing values out of one of the result files), these operations could be run on a single node. 
This could be added to the spec by changing the DockerExecutor parameter in the Task to be an array. The runner would then just call each of the docker executor statements in order all on the same machine, one after the other. So it wouldn't be a very difficult to implement.

geoffjentry:
@dionloy Hi Dion - from my perspective this API is intended for individual jobs, with a higher level engine (e.g. a cwl runner or Cromwell) orchestrating those. For instance perhaps one job is sent to google cloud, one to azure, one to aws, one to ibm - but all within the same workflow.

That said, @kellrott makes a good point - at some point we're going to need to crack the nut of combining tasks, streaming, etc. We've found it to not really be necessary on GCS although that's really been an artifact of the fast internal transfer rates, it's not likely to hold up on the other cloud services.

I like that notion of being able to pass in multiple requests intended to be run on a single machine.

dionloy:
@geoffjentry So, I see those two things as mutually exclusive, as you can support multiple requests to be run on a single machine via a repeated field for the command to be executed.  I do think that's a pretty good thing to support with the caveat that we should encourage people to make their execution stages as small as possible, to take advantage of pre-emptable VMs for significant cost savings. But as mentioned, this can be done without the dependsOn concept.

@kellrott I do agree that having the concept available would make it easier to implement work engines -- however, it comes at an implementation cost. It's a simple concept, but once implemented, you have to deal with issues like cycles; not to mention to be completely useful you'd want to also provide some abstraction for scatter / gather etc.. This will snowball into requiring each implementation to fully create a work engine themselves -- that's a significant barrier of entry for an implementor (I definitely don't want to be in that business =) ). 

Would making the cmd parameter an array to be executed on the same machine suffice in addressing the need instead of 'dependsOn' for you?

Also, the notion of network being slow on other cloud providers should be tested if we are making design decisions around that. I would find it hard to believe that Amazon and Azure do not have the same level of network vs "local" disk performance as Google Cloud. AFAIK network has outpaced local disk I/O for a few years now.

Thanks!

geoffjentry:
@dionloy I hadn't noticed the `repeated` - is that one of the diffs from the Google API? I hadn't seen that previously and I don't see it documented either there nor here.

On the network speed issue, my understanding is that there can be as much as a 4-5x decrease in data localization speed onto a VM using some of the other cloud vendors. Localizing/Delocalizing reference data and such would be a lot slower doing it the naive way on non-Google clouds.

Reference data isn't a great example because in your all's world one could mount a persistent disk as reference can be read-only but that also happens with large bams and such.

kellrott:
I think these two concepts, 'multipleTasks' and 'dependsOn', serve two different purposes. 

multipleTasks can be used for 'post task' jobs that the work engine needs to do. For example; after Galaxy is done running a user provided tool container, it needs to do meta-data generation for the resultant files (for example generate .bai files for the BAM outputs), or how WDL supports some output reading functions (like the read_map output function). The code that does those tasks isn't in the user container, so being able to tack on a few operations to do the engine specific jobs would make it a lot easier. (Later on we can open up the can of worms that would be allowing people to chain the stdout/stdin of multiple containers together on the same machine ;-)

dependsOn is about decreasing the amount of task polling that the work engine would need to do in order to move pipelines forward (and potentially allow task execution implementors to optimize systems to decrease file input/output staging). I think the scatter/gather example is a good one. The work engine would generate 1000 scatter tasks, get all of their IDs as they are created, and then generate the task of the gather step, and add in the dependsOn array of all the ids it collected. That last job would be delayed until all of its dependencies are ready, and the work engine would only have to watch that last task. Without it, the work engine would insert 1000 tasks, and then cycle asking status on them, wait until it gets 1000 'task complete' responses, and then insert the final task. I don't think it would induce cycles. The rule would be that the dependsOn array would be set of valid taskOp ids, you wouldn't be able to create a cycle because you can't insert a dependency on a taskOp that doesn't exist yet.

dionloy:
Kyle, good point about reducing the polling frequency via a gather
dependsOn. I'm gradually coming around to this idea, but I do still think
that it will complicate implementation which may impact the potential # of
implementors.

Jeff -- sorry, I meant a future change to use a repeated field there, it
does not currently exist =).

geoffjentry:
The use case which most interests me would be something like "Run these 3 tasks w/o having to push/pull the data back & forth from cloud storage in between" - the naive way would be to require them to be on the same VM in succession although if the request was made in a generic way it would allow a provider to implement in the most efficient mechanism they have (e.g. a shared disk). As @kellrott pointed out this would be potentially satisfiable with the `dependsOn` but I don't think it necessarily is as it could be set up such that it can only guarantee the limited polling concept

tetron:
@kellrott I'm with @dionloy, having dependencies in the task execution api just moves complexity around (or worse, duplicates it), it doesn't make it simpler.

@geoffgentry I agree, being able to chain or pipe tasks on the same node is an interesting feature.  Indeed, in CWL it should be possible for an optimizer to determine when to schedule tasks on the same node automatically (although no one had written any code to demonstrate that yet).

geoffjentry:
@tetron Thinking about it a bit more I think there are a few use cases in there.

One would be the reusing of the node - for whatever reason the execution engine decides this would be an optimization.

Another would be a way to specify to directly feed the outputs of one task into another task. This *could* be accomplished via reusing a node, but not necessarily. For instance in the google case it could be accomplished via a persistent disk. Having the directive be more generic would be good I think since it allows a vendor to implement it in the way which makes the most sense for them.

The third I'm thinking of would be along the lines of an SGE job array. As an example, if I'm using google and want to have a large scatter/gather which exceeds my quota, it'd be good to alert google that the 100000 jobs I just sent it are all the same shape so they can just keep reusing the VM. This isn't quite the same as the first one as it's allowing the underlying system make the optimization which it sees fit.

pgrosu:
I think the dependency approach has already been cracked via FlumeJava/Dataflow for the execution plan, where in our case it can mimic the intermediate-level MapShuffleCombineReduce (MSCR) operation with some slight changes where they go into wait states until the dependent results are available.  Below are links to the papers:

[FlumeJava](http://pages.cs.wisc.edu/~akella/CS838/F12/838-CloudPapers/FlumeJava.pdf)
[Dataflow](http://research.google.com/pubs/archive/43864.pdf)

Hope it helps,
~p

kellrott:
My read on the feedback so far is that I should probably go ahead with the 'multipleTasks' option, but skip the 'dependsOn' for now.

Are there any other issues that should be reviewed?
What will the voting for merge strategy be?

geoffjentry:
@kellrott Sounds good to me.

What about something like the apache model, we set a time N (2?) days in the future and say that anyone w/ objections needs to have spoken by then

kellrott:
I've made the 'repeated tasks' and added some inline documentation. 
How about we bring up a vote on the monday call?

geoffjentry:
Sounds good. I won't be at the call but this has a :+1: from me.

@dionloy and @rnpandya - can you all make sure this looks good from your end and/or join the call on monday?

tetron:
I realize this is a work in progress draft proposal, but is anyone volunteering to write a minimal open source reference implementation?

geoffjentry:
I *might* be able to do that, my calendar is a wee bit more open now that BioIT is in the books but I'm still catching up to a bunch of stuff I shirked along the way. If no one else volunteers I can have a better sense in a week or so

briandoconnor:
Who wants to make a very, very simple reference implementation?  @kellrott ?

briandoconnor:
Carlos asked about stderr/stdout/failure code etc

proto/task_execution.proto
```diff
@@ -0,0 +1,132 @@
...
+}
+
+message Resources {
+  uint32 minimumCpuCores = 1;
+  bool preemptible = 2;
+  uint32 minimumRamGb = 3;
+  repeated Disk disks = 4;
+  repeated string zones = 5;
+}
+
```
ccarrizo: would it be useful to add in this field a reference to the execution context?
e.g. if we want to link the results of one task to another, instead of forcing the implementation to re-use a node, it might be more efficient to define an execution context (e.g. as an output), and then require it as resource on the second task, thus the implementing part might decide if this means using the same node or sharing the same disk or whatever implementation fits their infrastructure.

proto/task_execution.proto
```diff
@@ -0,0 +1,132 @@
...
+  bool autoDelete = 4;    /// Automatically delete after usage
+  bool readOnly = 5;      /// mount into docker as read only
+  string mountPoint = 6;  /// mount point for disk inside the docker container
+}
+
+message Resources {
+  uint32 minimumCpuCores = 1;
+  bool preemptible = 2;
+  uint32 minimumRamGb = 3;
+  repeated Disk disks = 4;
```
ccarrizo: I think HW requirements should be a richer and more flexible argument, 
As examples:
-A developer might have identified what is the minCPUCores, but also might know what is the scaling policy (e.g. a max value or not-bounded).
-It might also be the case the developer can request for specific HW and arguments: if x.available | args+" --hw=x".
This way the developers can give hints to the platform implementer to make the best possible use of the Sw on any given HW

proto/task_execution.proto
```diff
@@ -0,0 +1,132 @@
...
+  repeated string zones = 5;
+}
+
+/// The description of a task to be run
+message Task {
+  string name = 1;         /// user name for task
+  string projectId = 2;    /// parameter for execution engine to define/store group information
+  string description = 3;  /// free text description of task
+  repeated FileParameter inputParameters = 4;  /// Files to be copied into system before tasks
```
ccarrizo: might this be a more generic input? e.g if a port or a pipe is specified as input

proto/task_execution.proto
geoffjentry: @ccarrizo Do you have suggestions on the concrete format?

I like the idea on paper but IMO something like this needs to be simple either via a) fairly limited syntax or b) very generic syntax which allows for arbitrary constructs

I think 'a' is better here so I'd want to see what the extra syntax would look like.

proto/task_execution.proto
geoffjentry: @ccarrizo My take here is that this isn't the final version and that piping/streaming hasn't really been discussed yet - although everyone participating knows it should exist. IMO it shouldn't tie up merging this in and then we could address piping in a future PR.

kellrott: My take on this is that inputs and outputs of the task are everything that has to be shipped in and out of the node prior to calling the Docker execution commands. The stdin/stdout piping of the actual execution commands is handled once everything is in place. I think that bit needs to be much more fleshed out, but that would probably be by adding more options to the DockerExecutor structure. I would like to add in logic there that would allow for chaining of the different DockerExecutor in the array together, so that the stdout of one DockerExecutor would be the stdin of another.
I'm hoping we can settle on this initial framework, and then start fleshing out those higher level details.

geoffjentry:
@kellrott Was a path for merging discussed at the call on monday? I agree with your last statement, it'd be good to get something in place and then start iterating on that.

geoffjentry:
I'm merging this as per our conversation on Monday.

PR #6, opened 2016-07-10
Adding method call for getting service configuration information.

kellrott:
Adding method to allow the client to obtain service configuration information. For the moment it is presented as a set of key-value pairs, with no standardized or required fields.

### DIFF
--- a/proto/task_execution.proto
+++ b/proto/task_execution.proto
@@ -161,8 +161,28 @@ message Job {
   repeated JobLog logs = 5;
 }
 
+//Blank request message for service request
+message ServiceInfoRequest {}
+
+//Information about Task Execution Service
+//May include information related (but not limited to)
+//resource availability and storage system information
+message ServiceInfo {
+  //System specific key/value pairs
+  //Example for a shared file system based storage system:
+  //storageType=sharedFile, baseDir=/path/to/shared/directory
+  map<string,string> metadata = 1;
+}
+
 //Web service to get, create, list and delete Tasks
 service TaskService {
+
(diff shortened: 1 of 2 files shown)


geoffjentry:
Could you give some examples of how you see this being used?

kellrott:
The primary usage right now is relaying configuration about the storage system that the workers will be using. In the reference implementation, there is an 'object store' that is directory on a shared file system. Testing scripts can find the 'object store' to copy files in with:

```
$ curl http://localhost:8000/v1/jobs-service
{"metadata":{"baseDir":"/Users/kellrott/workspaces/task-execution-reference/storage","storageType":"sharedFile"}}
```

Once the storage system become a real object store, like a swift S3 store, you'll have to pass in other configs, like the endpoint URL. For implementing this with Arvados, there would have to be config information for talking to the keep server.

I know its a bit messy to put the configs into a key/value map, rather then having a fixed schema, but we've already said that we're going to leave the linked storage system up to the implementor, and there is no way to predict the kind of data they would provide.

kellrott:
One things I could do to clean it up would be to create a specific 'storageConfig' field that is a map to store the storage specific configuration info, rather then having a generic 'metadata' catch-all. 
Usage stats and resource values could go into other fields.

kellrott:
I've changed the metadata structure to be 'storageConfig'. Other fields for metadata can be adding in later.

kellrott:
The reference server requires this changed, and because this proposal has been sitting for 2 weeks, without any disapprovals, I'm going to go ahead and merge it.

PR #8, opened 2016-08-19
Adding workdir parameter to the DockerExecutor message

kellrott:
Adding a workdir file so that users can fill out the '--workdir' flag in the docker command and define which directory the command will be executed in.

### DIFF
--- a/proto/task_execution.proto
+++ b/proto/task_execution.proto
@@ -23,6 +23,11 @@ message TaskParameter {
   //data is a directory, if used for an output all the files in the directory
   //will be copied to the storage location
   bool   directory = 5;
+  //if the parameter is an output, should the element be created before executing
+  //the command. For example if saving the working directory as an output,
+  //the directory needs to exist before the command is called. If false, it is
+  //assumed that the user will create the element as a part of the operation
+  bool   create = 6;
 }
 
 //A command line to be executed and the docker container to run it
--- a/swagger/proto/task_execution.swagger.json
+++ b/swagger/proto/task_execution.swagger.json
@@ -395,6 +395,11 @@
     "ga4gh_task_execTaskParameter": {
       "type": "object",
       "properties": {
+        "create": {
+          "type": "boolean",
+          "format": "boolean",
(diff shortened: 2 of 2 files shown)


kellrott:
If there is no objection, I'll go ahead and merge this.

geoffjentry:
👍

proto/task_execution.proto
```diff
@@ -23,6 +23,11 @@ message TaskParameter {
   //data is a directory, if used for an output all the files in the directory
   //will be copied to the storage location
   bool   directory = 5;
+  //if the parameter is an output, should the element be created before executing
```
pgrosu: Sorry, but just trying to clarify for myself, what is the difference between a `location`, a `path` and a `directory` and how should they be used?  And where are `data` and volumes defined?

```Proto
  //location in long term storage, is a url specific to the implementing 
  //system. For example s3://my-object-store/file1 or gs://my-bucket/file2 or
  //file:///path/to/my/file
  string location = 3;
  //path in the machine file system. Note, this MUST be a path that exists
  //within one of the defined volumes
  string path = 4;
  //data is a directory, if used for an output all the files in the directory
  //will be copied to the storage location
  bool directory = 5;
```

Thanks,
~p

kellrott: A 'location' is where the file is stored long term, ie an object store url like s3://my-bucket/the_file.txt. The path is there the file is going to be inside the container running on the machine (thus it must be a subpath of one of the declared volumes attached to the container). directory is a boolean hint that these things are referring to a directory, so implementations can do recursive copying and in the case that it is an output directory and the create bool is also true, the system will go ahead an create the directory before running the command.

pgrosu: Thanks for clarifying.  My Spidey sense is telling me that folks might misinterpret `path`.  Maybe in a future PR possibly re-naming it to `containerMappedPath`, or just `containerPath` might be more explicit.  The location could be named as `cloudPath` or something like that, with directory renamed as `arePathsDirectories` - though many times an API can tell if a path is a directory or a file.

On a separate nit, some of the documentation for the fields seem like phrases and start with lower-case letters.  As this will become a standard that others will try to implement, it might be nice to have them as "proper" as possible.  For example, I had to read this sentence about five times before it clicked:

>//if the parameter is an output, should the element be created before executing
//the command.

PR #20, opened 2016-12-16
Change jobListResponse to show JobID, not Job

jeenalee:
This patch changes `jobListResponse` to include `JobID`s, instead of `Jobs`. This will make the returned value to be much smaller.

### DIFF
--- a/proto/task_execution.proto
+++ b/proto/task_execution.proto
@@ -121,10 +121,14 @@ message JobListRequest {
   string pageToken = 4;
 }
 
+message JobDesc {
+  string jobID  = 1;
+  State state = 2;
+}
+
 message JobListResponse {
-  repeated string jobID  = 1;
-  repeated State state = 2;
-  string nextPageToken = 3;
+   repeated JobDesc jobs = 1;
+   string nextPageToken = 2;
 }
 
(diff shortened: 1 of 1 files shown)


kellrott:
The message 
```
message JobListResponse {
   repeated string jobID  = 1;
   repeated State state = 2;
   string nextPageToken = 3;
}
```
would end up looking like 

```
{
"jobID" : [ "abc1", "abc2", "abc3" ],
"state" : ["error", "running", "queued"]
}
```
That seems less intuitive then something like 

```
message JobDesc {
  string jobID  = 1;
  State state = 2;
}
message JobListResponse {
   repeated JobDesc jobs = 1;
   string nextPageToken = 2;
}
```

which would produce a message like


```
{
"jobs" : [ 
   {"jobID" : "abc1", "state" : "error"}, 
   {"jobID" : "abc2", "state" : "running"}, 
   {"jobID" : "abc3", "state" : "queued"} 
]}
```
That may be more verbose, but it is a bit more 'natural'.

kellrott:
Looks good, reduces verbosity of default listing and additional fields can be added into JobDesc as needed.

No downvotes, so merging.

PR #21, opened 2017-01-19
Taskdesc

kellrott:
Adds small task description data structure when listing out job information.

### DIFF
--- a/proto/task_execution.proto
+++ b/proto/task_execution.proto
@@ -142,12 +142,27 @@ message JobListRequest {
   string pageToken = 4;
 }
 
+// Small description of tasks, returned by server during listing
+message TaskDesc {
+  //OPTIONAL
+  //user name for task
+  string name = 1;
+  //OPTIONAL
+  //parameter for execution engine to define/store group information
+  string projectID = 2;
+  //OPTIONAL
+  //free text description of task
+  string description = 3;
+}
+
 //Small description of jobs, returned by server during listing
 message JobDesc {
   //REQUIRED
   string jobID  = 1;
   //REQUIRED
   State state = 2;
+  // REQUIRED short description of task
+  TaskDesc task = 3;
 }
 
 //Return envelope


geoffjentry:
What's the motivating use case for this?

kellrott:
Improvements to the HTML display: https://github.com/bmeg/task-execution-server/pull/55
Primarily, this additional information provides enough information to describe the task when the user is presented with a list of all jobs.

geoffjentry:
Shouldn't that be tracked by the outer service (in your case, also named TES)? 

For instance when we interact with JES we're just given opaque operation IDs and that's fine as users are interacting with us and not JES. I don't see why this would need to be different.

kellrott:
I think access to the job log, and historical information is important. From the larger perspective (not just this PR) In working on some larger workflow projects, being able to retrospect/search all of the jobs was incredibly important. 

I think long term, I'd like to propose a second 'service' definition in the specification. The `TaskService` defines 5 endpoints that must be operational to comply with the core schema. I think we should define a second `TaskSearchService` that add additional endpoints that can be implemented for 'extended compliance'. This would include things like being able to search for previously run jobs based on the inputs used or outputs created, which docker image or command line was used. 

We can then define levels of compliance for different implementations, with core compliance being just doing the `TaskService` and extended compliance implementing the `TaskSearchService` as well.

geoffjentry:
I agree that this is useful information but disagree that this API is the correct location for it. I think the task API should be as lean as possible and this feels like extra bells and whistles which could/should be maintained by whatever software is talking to it.

IMO a good litmus test for TES is "do JES and Batch support this?" as (again IMO) TES should be a generic equivalent to those APIs.

I feel that this is conflating TES-the-server with TES-the-API and that this belongs in the former.

geoffjentry:
@kellrott Wasn't sure if this is the PR you mentioned that you might have swayed me over or not :)

This is the one open PR that I'm still opposed to, all of the rest seem good to me.

kellrott:
This PR is mainly driven by the need to have 'some but not all' the information about a set of jobs. At first, the job list was all of the information (including logs and command lines) which could become onerous when trying get info on 10K jobs. But the second version, ie just the ID, requires another request for every job to get general info (like job name and project). 
This is just trying to find a way to split the difference.

Part of the needs for this PR may be addressed in future plans. We've been talking about adding an additional endpoint, something like `/v1/jobs-search` that allows for different ways about querying the jobs. Stuff like search for jobs by docker container or input/output files. It would be an addition to the spec, and considered optional for compliance.

buchanae:
Looks like I committed these changes and I don't even understand the usefulness now. The Task structure doesn't contain any logs and so should be lightweight as is. I suggest we close this PR.

kellrott:
Closing in favor of another method

PR #25, opened 2017-02-06
Output File listing

kellrott:
This adds a record of all output files (and file size) to job record. This expands on the `outputs` TaskParameter as that is a request, that can include base directories, and this is the manifest of what was actually done including all files contained in those directories.

Many workflow engines need to 'glob' the output directories to see what files were created (as some of the file names are not predetermined). Without this listing, it is incumbent on the workflow engine to independently contact the object store and get the listing manually. Having this listing removes the need for that.

### DIFF
--- a/proto/task_execution.proto
+++ b/proto/task_execution.proto
@@ -221,6 +221,24 @@ message JobLog {
   repeated Ports ports = 8;
 }
 
+
+//Log of file output by workflow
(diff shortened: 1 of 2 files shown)


kellrott:
Dropping in favor of #26 , which is a cherry-picked patch of the relevant changes.

PR #27, opened 2017-02-10
Remove JobID message type

buchanae:
Removing the JobID message cleans up the messages and related code a bit. Schema clients previously needed code like:

```
jobid := JobID.Value
```

...which I didn't like looking at :) 

This also allows the request/response message types to grow in the future, without worrying about whether that field belongs on a "JobID" message. And it's a more consistent pattern. Most RPC endpoints have an EndpointNameRequest message and an EnpointNameResponse message.

### DIFF
--- a/proto/task_execution.proto
+++ b/proto/task_execution.proto
@@ -156,11 +156,6 @@ message JobListResponse {
    string nextPageToken = 2;
 }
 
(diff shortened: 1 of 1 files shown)


geoffjentry:
👍

PR #35, opened 2017-03-30
Remove (or clarify) TaskParameter.create

buchanae:
Possibly, the `TaskParameter.create` field is not needed. If an output directory needs to exist, that can easily (and more consistently) be done with volumes. I don't think there's a case where an output file needs to be created, so this is directory specific.

If there's a good use case for this field, then I'd at least like to get that case clearly documented.


geoffjentry:
👍

kellrott:
Managed in #50

Issue #11, opened 2016-10-21
Add 'contents' to TaskParameter


kellrott:
Allow for small scripts to be injected into the system, without having to go through the object store, using a 'contents' field in the TaskParameter (and leaving the location field blank)

Example:

```
{ 
   "inputs" : [
       { "path" : "/tmp/runtime.sh", 
          "contents" : "#!/bin/bash\necho \"What up dog?\"\n"
       }
    ],
    "docker" : [
        {  
            "imageName" : "ubuntu",
        "cmd" : ["/bin/bash", "/tmp/runtime.sh"]
        }
    ]
}
```

briandoconnor:
-1 I don't think I'd use this... I'm most concerned about large file inputs which I have to upload to an object store.  So I can't really avoid using an object store and would just bake in scripts into my docker image.

buchanae:
I'm 50/50 on this.

I feel like client libraries will already have code to support communication with a filesystem (e.g. object storage), so this probably isn't absolutely necessary and would add one more item to the list of things needed to implement a TES server.

Also, sometimes it's nice for debugging when files like this are written to an accessible filesystem, so that you can go hunting around afterwards to understand what happened. If they are written only to the worker node, you might have less/no access to those files for debugging.

Possibly this adds security risk as well, since we're talking about writing a file that's going to get executed and we need to apply limits to the size.

But, on the other hand, it would possibly be easy to implement and might be handy for keeping temporary files out of the object store. And, you might need just as much logic to figure out how/where/when you're going to store/cleanup run scripts in the object store.

buchanae:
This made me think of extensions:

Some APIs allow for extensions (I think Kubernetes does this) which is often a useful way to (dis)prove the utility of features before accepting into the core. Might be something keep in mind for the longer future of TES.

geoffjentry:
I wouldn't object to this but I also feel it's necessary either. Call that less "50/50" and more "I don't care enough to argue either way" :)

buchanae:
I've been chewing on a related idea, which is to give TES a limited storage API which would include get, put, and stat operations. TES already has a concept of storage backends, configuration, and URLs such as `file:///path/to/foo` and `s3://bucket/path/to/foo`.

`get` and `put` are obvious, I presume. `stat` would return metadata about the file.

[...]

PR #41, opened 2017-03-31
TaskView

buchanae:
One concern I've had is that job stdout/stderr logs will impose a heavy network load on both clients and servers. For example, a user refreshes a dashboard list of 100 tasks, and each task includes 10K of logs. I've been concerned about the #11 (contents parameter) for the same reason.

This PR introduces a TaskView enum and corresponding field on the ListTasksRequest (aka JobListRequest) message, which allows a client to switch between the BASIC and FULL views of a task.

Additionally, this removes the TaskDesc message, which is a nice benefit IMO.

This was inspired by Google's API design guide:
https://cloud.google.com/apis/design/design_patterns#resource_view

This branch is based on the branch from #39 so a cleaner comparison is here:
https://github.com/buchanae/task-execution-schemas/compare/style...buchanae:full-view

### DIFF
--- a/proto/task_execution.proto
+++ b/proto/task_execution.proto
@@ -337,6 +337,27 @@ message ListTasksRequest {
   // If unspecified, returns the first page of results.
   // See ListTasksResponse.next_page_token
   string page_token = 4;
+
+  // OPTIONAL
+  //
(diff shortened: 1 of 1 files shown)


geoffjentry:
Seems reasonable 👍

PR #51, opened 2017-04-11
Redefine ServiceInfo fields

### LINKED ISSUE #28: Expand ServiceInfo message
The ServiceInfo message currently contains only a map<string,string> for storage configuration, but storage configuration is likely more complex than a simple string to string map. For reference:

```proto
//Information about Task Execution Service
//May include information related (but not limited to)
//resource availability and storage system information
message ServiceInfo {
  //System specific key/value pairs
  //Example for a shared file system based storage system:
  //storageType=sharedFile, baseDir=/path/to/shared/directory
  map<string,string> storageConfig = 1;
}
```

Funnel's storage configuration currently includes (Go code):
```go
type LocalStorage struct {
	AllowedDirs []string
}

type S3Storage struct {
	Endpoint string
	Key      string
	Secret   string
}
```

Packing these into a map<string,string> requires some trick, such as a composite key such as "Local.AllowedDirs". Having to define custom string packing/unpacking when we're using protobuf everywhere else feels wrong; it adds complexity and therefore a source of bugs and confusion.

buchanae:
Closes #28

### DIFF
--- a/proto/task_execution.proto
+++ b/proto/task_execution.proto
@@ -430,12 +430,20 @@ message ServiceInfoRequest {}
 // and other documentation.
 message ServiceInfo {
 
-  // OPTIONAL
+  // Returns the name of the service, e.g. "ohsu-compbio-funnel".
+  string name = 1;
+
+  // Returns a documentation string, e.g. "Hey, we're OHSU Comp. Bio!".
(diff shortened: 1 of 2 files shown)


kellrott:
Got +1 in #28

PR #71, opened 2017-06-09
TaskParameter.contents -> content

buchanae:


prismofeverything:
+1 This makes more sense to me.

briandoconnor:
Seems fine to me...

task_execution.proto
```diff
@@ -113,7 +113,7 @@ message TaskParameter {
   // UTF-8 encoded
   //
   // If contents is not empty, "url" must be ignored.
```
geoffjentry: should update the comment :)

geoffjentry:
:+1:

task_execution.proto
buchanae: Good catch. Updated.

pditommaso:
👍

Issue #73, opened 2017-06-15
Service logs


buchanae:
While there are many places for logs from the task/executor, there is no place for logs from the TES service. There may error/info logs which would be nice to convey to the client:

- system errors (problem with TES service)
- task errors only found during execution (bad executor container image name)

buchanae:
Possibly this could fit in `TaskLog.Metadata`

buchanae:
Proposed `TaskLog.Syslogs` in #80

PR #80, opened 2017-10-05
System logs field

buchanae:
In Funnel, we've found that there are a lot of logs coming from the system and not the executor (container) that are important: an error message from the object store client, an error message from the system about a disk being full, a docker run command, etc. We've been avoiding adding this to TES and putting it in separate endpoints, or accessing these logs via external systems such as rsyslog or Kibana, but I think it's worth proposing here. It's an optional field, and in the logs message, so not much overhead to implementations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,7 +107,7 @@ The return value will be a minimal description of the task state.
 To get more information, you can change the task view using the `view` URL query parameter.  
 
 The `basic` view will include all task fields except a few which might be
-large strings (stdout/err logging, input parameter contents).
+large strings (stdout/err/system logging, input parameter contents).
 
 ```HTTP
 GET /v1/tasks/task-1234?view=BASIC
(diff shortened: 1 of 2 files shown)


buchanae:
Closes #73

briandoconnor:
Do you need STDERR and STDOUT?  I like the idea of collecting these and making them accessible for the reasons you say.  You may need to specify in the comments of the schema that this may be subject to a line limit for a given implementor:

http://editor.swagger.io/#/?import=https://raw.githubusercontent.com/ga4gh/workflow-execution-schemas/0.1.0/swagger/proto/workflow_execution.swagger.json

buchanae:
> Do you need STDERR and STDOUT? I like the idea of collecting these and making them accessible for the reasons you say. You may need to specify in the comments of the schema that this may be subject to a line limit for a given implementor:

These logs are not tied to any single process, so stdout/err don't fit. These are any logs the system decides are relevant outside of the actual executor/docker process. I've improved the field docs in the latest commit, adding some notes about the (lack of) specification around the format/size/etc.

geoffjentry:
👍

pditommaso:
👍

buchanae:
I've added documentation to state that the system logs are only available in the `FULL` task view, which is consistent with task stdout/err logs and other fields which are potentially more than a small string.

PR #81, opened 2017-10-09
Fix python syntax

giannidallatorre:


buchanae:
I think that's JSON, not python, so "false" would be correct.

giannidallatorre:
Hi Alex,

Oh sorry, you are right I was testing an implantation in Python and I got confused, my bad.

Actually the comma at the end line, of the last element of the lists is not JSON compliant, while it would be with Python!

Cheers,
Gianni

buchanae:
Good point about the commas, I'll run these through a JSON linter. Thanks!

PR #83, opened 2017-10-15
suggestions for more consistent field names in Executor

prismofeverything:
Going through the schema and implementing a remote interface for funnel, all of the field names are so concise and elegant that these three field names stuck out.

I am proposing three changes:

* image_name --> image

Most of the other fields throughout the schema are single names, and the word `image` does not need to be disambiguated with any other fields in the schema, so this seems like the natural choice.

* cmd --> command

The schema is refreshingly light on abbreviated names to the point where this one stands out. Why not just `command`?

* environ --> env

This one is funny because it is really just a different abbreviation of `environment`. I would almost prefer `environment` since it is nice to be clear, but we have such a long-standing tradition of using the three letters ENV to stand for the environment (even the bash program called `env`) that I think this one makes sense.

Open to all thoughts and ideas on this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,8 +46,8 @@ an MD5 checksum on an input file and uploads the output:
     },
     "executors" : [
       {
-        "image_name" : "ubuntu",
-        "cmd" : ["md5sum", "/container/input"],
+        "image" : "ubuntu",
+        "command" : ["md5sum", "/container/input"],
         "stdout" : "/container/output",
         "stderr" : "/container/stderr",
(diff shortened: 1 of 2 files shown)


briandoconnor:
Changes seems fine to me

geoffjentry:
:+1: this all seems right

pditommaso:
👍

buchanae:
@prismofeverything can you update the README as well please?

prismofeverything:
Okay, done.

PR #86, opened 2017-10-23
Rename TaskParameter -> Input and Output

buchanae:
Another minor name cleanup, this splits `TaskParameter` to `Input` and `Output`, which are simpler, more obvious names. It also removes the `contents` field from `Output`, which didn't exactly make sense for a `TaskParameter` used in the `Task.Outputs` field.


geoffjentry:
👍

pditommaso:
👍

PR #89, opened 2017-10-23
Remove paused (or else define it more clearly)

buchanae:
This seems like a poorly defined carryover from the Google Genomics API. There's little/no explanation of the behavior of pausing. Can a user pause/unpause? Is pausing controlled solely by the system? If so, does the user need to know that the task is paused?

Not necessarily against the concept of pausing, just trying to cut out the fuzzy parts of the spec until they're clearly defined.

### DIFF
--- a/task_execution.proto
+++ b/task_execution.proto
@@ -329,9 +329,6 @@ enum State {
   QUEUED = 1;
   INITIALIZING = 2;
   RUNNING = 3;
-  // An implementation *may* have the ability to pause a task,
-  // but this is not required.
-  PAUSED = 4;
   COMPLETE = 5;
   ERROR = 6;
   SYSTEM_ERROR = 7;


geoffjentry:
Don't think this came from PAPI, not sure where it came from. :+1:

pditommaso:
Several batch schedulers implements a suspended/hold state for tasks, but I agree that adds no value to expose it to client. 👍

buchanae:
@pditommaso Do you have some examples? I'd like to do some research on those.

geoffjentry:
@buchanae for instance i know LSF has it, not sure about SGE. It definitely exists in HPC setups but I question if we should be bringing it forward, the world that TES would be sitting on top of is likely to be a lot more heterogenous than the HPC world (which are all pretty similar once one squints and takes a step back)

pditommaso:
@buchanae yes.

* LSF has `PSUSP`, `SSUSP` and `USUSP` hold statuses.
* HTCConder: `H` status 
* SLURM: `S` (suspended) status
* PBS: `S` and `H` status 
* SGE: `hqw` status. 

Sorry I don't have a proper link to the documentation of each of them. I getting from the NF source code.

buchanae:
Thanks guys. I can imagine, as a user, wanting to know what my task isn't currently processing and is waiting. For example, if someone asks me how much longer the batch of tasks should take and I know the average runtime, I'll give the wrong estimation if I don't know that some of the tasks are currently paused.

pditommaso:
Yes, it could be useful in that way. I'm neutral on this.

buchanae:
This was accidentally merged, but was reverted by another commit. Sounds like we're not sure if we want to remove it. I've created #99 for any further discussion. Sorry for the confusion.

To be clear, the paused state is still in the spec.

PR #90, opened 2017-10-23
Add creation_time field

buchanae:
`create_time` describes when the task was created, which is very useful for finding tasks later, and for sorting tasks in a dashboard.

I went with `create_time` in order to by consistent with the `start_time` and `end_time` naming.

### DIFF
--- a/task_execution.proto
+++ b/task_execution.proto
@@ -69,6 +69,12 @@ message Task {
   // Normally, this will contain only one entry, but in the case where
   // a task fails and is retried, an entry will be appended to this list.
   repeated TaskLog logs = 12;
+
+  // OUTPUT ONLY, REQUIRED
+  //
+  // Date + time the task was created, in RFC 3339 format.
+  // This is set by the system, not the client.
+  string creation_time = 13;
 }
 
 enum FileType {


task_execution.proto
```diff
@@ -69,6 +69,12 @@ message Task {
   // Normally, this will contain only one entry, but in the case where
   // a task fails and is retried, an entry will be appended to this list.
   repeated TaskLog logs = 12;
+
+  // OUTPUT ONLY, REQUIRED
+  //
+  // Date + time the task was created, in RFC 3339 format.
+  // This is set by the system, not the client.
+  string create_time = 13;
```
geoffjentry: `creation_time`?

geoffjentry:
👍

task_execution.proto
buchanae: Ya, `create_time` has thrown a few people off now. If we're less consistent about the `_time` suffix, then `created` or `created_at` are good options too.

pditommaso:
👍 (`creation_time` sounds better)

buchanae:
Changed to `creation_time`

Issue #84, opened 2017-10-17
Define custom resource request schema


buchanae:
There's no place to request some types of resources:
- GPUs
- Disk type
- CPU architecture
- AWS Spot Market bidding strategy
- Cloud/cluster-specific resources such as specialized hardware (e.g. Tensor Processing Unit)

Also, some existing concepts in the `Resources` schema are not clear
- zones: not a consistent concept across clusters, e.g. in GCE there are multiple global regions, each with multiple zones.
- `preemptible` does not fully capture the concept in AWS

A proposed solution is to remove some concepts from the schema, such as `preemptible` and `zones`, and add a new field of type `map[string]string` to allow more flexible, implementation-specific resource requests.

pditommaso:
My feeling is that this kind of requests (eg. CPU architecture, spot instance, bidding strategy, etc) should be delegated to a `queue` concept which could be (optionally) specified by a task.

geoffjentry:
Something we've run into over time is that there's a tension between a desire for generic key/values that an execution backend can feel free to interpret and people's desire for something to work the same no matter where they go. This is less bad for concepts which don't map cleanly at all but once you start getting to things which are semi-shared it becomes problematic.

If I had a good solution to this problem I'd have solved it already, but worth pointing out.

buchanae:
> My feeling is that this kind of requests (eg. CPU architecture, spot instance, bidding strategy, etc) should be delegated to a queue concept which could be (optionally) specified by a task.

If using queue names with a large workflow, would each step need to define its queue? If you moved that workflow to another cluster, would you have to edit all those queue names?

I guess, alternatively, one might argue that it's the workflow engine's responsibility to map workflow task resource requests into TES queue names. That has some tradeoffs:

- requires administrators to manage a set of queues for all types of tasks.
- requires the workflow engine to know details about the TES service's resources, which makes it more difficult to point a workflow engine at an arbitrary TES service.
- adds an extra indirection for understanding what a task actually requested, which adds a little more work to, for example, auditing a history of tasks to understand what resources were used.
- not all tasks are driven by an intelligent workflow engine. Ad hoc clients would need to know how to use queues effectively too.

> Something we've run into over time is that there's a tension between a desire for generic key/values that an execution backend can feel free to interpret and people's desire for something to work the same no matter where they go. This is less bad for concepts which don't map cleanly at all but once you start getting to things which are semi-shared it becomes problematic.

[...]

PR #91, opened 2017-10-23
Remove project fields

buchanae:
The `project` field is underspecified – is it required to be a cloud project ID? Or just more metadata? Or both? It could be defined via the `tags` field, which would make it simpler to filter via `ListTasks` and might fit to #84 which uses `tags` for cloud/deployment  specific resource requests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,6 @@ an MD5 checksum on an input file and uploads the output:
 {
     "name":        "MD5 example",
     "description": "Task which runs md5sum on the input file.",
-    "project":     "tes-example-project-id",
     "tags": {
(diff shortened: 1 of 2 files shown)


geoffjentry:
FWIW for a system we're putting on top of Cromwell we went with the tag route. We found that `project` was both too limiting as well as too generically named, as you note yourself.

We decided that systems sitting on top of Cromwell could set tags to their hearts content and work that way and that way clients could have whatever they want.

geoffjentry:
tl;dr that's a :+1: for tags instead of project

pditommaso:
👍

PR #94, opened 2017-10-24
Clarify Executor.stdout/err docs and implementation requirements.

buchanae:
While developing Funnel, we've found that ensuring that these logs are reliably the correct tail of the content is tricky and feels like an unnecessary burden to implementors. I'm proposing we remove the requirement of the tail, and free up implementations to do what works best for them and their users. There is a fallback of writing the stdout/err to a file and uploading that file to permanent storage, as a way to consistently guarantee you get the full content.

### DIFF
--- a/task_execution.proto
+++ b/task_execution.proto
@@ -267,16 +267,28 @@ message ExecutorLog {
 
   // OPTIONAL
   //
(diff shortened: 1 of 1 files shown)


pditommaso:
👍

Issue #79, opened 2017-10-02
Volume specification unclear


erikvdbergh:
Currently there is a volume specification allowed in a task definition as follows:
```
repeated string volumes = 10;
```
It is also stated that these volumes should be shared between executors.
It is not clear to me how the volumes would be specified. Will this just be a directory on the container that should be mounted as the same directory in each executor? And does the 'size_gb' parameter link to these volumes? What amount of space should be reserved for them?

Perhaps more detailed Volume objects are required e.g.:
```
"volumes": [
  { 
    "name": "My first volume",
    "path": "/container/vol1",
    "size_gb": 20.0
  },
  { 
    "name": "My second volume",
    "path": "/container/vol2",
    "size_gb": 30.0
  }
]
```

I see in the mailing list that the initial idea for TES included a similar declaration. In the Funnel implementation there is an AddVolume method (https://github.com/ohsu-comp-bio/funnel/blob/master/worker/file_mapper.go#L88), but it is never called anywhere. It also defines a `readonly` parameter.

Curious to hear your thoughts.

buchanae:
I agree, it's not documented well, and has confused multiple people.

Volumes are intended to allow a directory to be shared between multiple executors. It does not have a specific size, nor does it relate to any particular storage location (such as mounting a disk in Google Cloud or AWS EBS).

Previously, each volume had a size, but it seemed to make resource requests more tedious: why describe each individual volume size when you can just request one set of resources for the whole task? So the "size_gb" is the disk size requested for the whole task. That changed in https://github.com/ga4gh/task-execution-schemas/pull/47

AddVolume does get called, internally by FileMapper, here: https://github.com/ohsu-comp-bio/funnel/blob/master/worker/file_mapper.go#L215

erikvdbergh:
Ok, so in practice, a volume entry will consist of a path, and the implementation must make sure that this path is available and shared between executors. Is that correct?

buchanae:
Correct.

Also, in the case of docker, that volume should be declared as a writeable volume mount.

Note, #64 asks similar questions.

buchanae:
@erikvdbergh I think we've clarified volumes by simplifying the spec and writing better documentation.

Is there anything left to do here? If not, can we close this?

delagoya:
Closing issue, as PR #95 was merged and no further comments made.

Issue #64, opened 2017-04-28
Request for clarification: Volumes field


mbookman:
The task_execution.proto indicates that tasks can contain Volumes, but it is not clear how this field should be used. Can this be expanded:

```
  // Declared volumes.
  // Volumes are shared between executors. Volumes for inputs and outputs are 
  // inferred and should not be delcared here.
  repeated string volumes = 10;
```
(Note there is a typo: delcared.)

What kind of string should one put here? Is it environment-specific?
For Google Cloud, for example, can I specify that a disk needs to be created of a certain size? Is this a mechanism for mounting an existing disk read-only?
Is this intended for a Grid Engine environment where an NFS volume needs to be mounted?

Thanks!

buchanae:
#47 and #30 are related.

As I understand, a volume string is a path inside the container, and the path on the host system is implementation-specific. For example, `volumes = ["/path/to/volume"]` would translate to `docker run -v /path/to/volume`.

Do these volumes exist primarily to share data between executors?

[Awhile back](https://github.com/ga4gh/task-execution-schemas/blob/a8272de6899bdbf15dc002198412848af55a13b2/proto/task_execution.proto#L81), Volume was a message and included `sizeGb` and `source`, which I think mapped more directly to cloud concepts.

Currently, a single disk size requirement lives in the [Resources](https://github.com/ga4gh/task-execution-schemas/blob/master/task_execution.proto#L183) message. I think the idea was that this is more simple for non-cloud (e.g. Grid Engine + NFS) environments, but maybe we strayed too far here and should rethink this?

buchanae:
@mbookman I think we've cleared up some of the confusion around volumes by simplifying the spec and adding better documentation.

If there's nothing left to do here, can you close this issue? Or if there's more left to do, can you clarify please? Thanks.

mbookman:
Sure. I think all of my issues with naming still apply, but I can't point to a clearly better replacement.

As a note, in Pipelines API v2alpha1, the analog is the "path" field inside the "mount" element.
When porting `dsub` to use v2alpha1, I similarly just needed to read the documentation to clarify.

I think the confusion comes naturally from the fact that a "volume" or "mount" without clarification can apply to a physical machine, a virtual machine, or a docker container. Your probably just going to need documentation to clarify, unless you name it `executor_volume`. Even then, you'll still need some amount of documentation about the lifespan of the volume.

Issue #59, opened 2017-04-20
Proposal: Remove stacked executors


geoffjentry:
There was discussion on the mailing list about removing (or rebranding) the stacked executor functionality. Creating an issue to allow for further discussion here.

buchanae:
Related mailing list discussions:
https://groups.google.com/a/genomicsandhealth.org/forum/#!topic/ga4gh-dwg-containers-workflows/qFq_jgoRCvs
https://groups.google.com/a/genomicsandhealth.org/forum/#!topic/ga4gh-dwg-containers-workflows/ccaqjysBvZY

buchanae:
To summarize a couple points from the mailing list:

1. There are concerns that stacked executors don't map well to concepts in Google and Amazon APIs, and implementing stacked executors in those systems could be difficult (or impossible?).
2. The original motivation for stacked executors was to provide initialization and finalization steps, rather than to encourage optimization of workflow steps.

tetron:
+1 on removing stacked executors.  There are workarounds for initialzation/finalization that don't involve multiple containers.

buchanae:
@tetron If you could give some short description of those workarounds, it might help the discussion. Please, thanks!

delagoya:
@buchanae @tetron There were several options discussed on the call. Basically it boils down to the  implementation detail of the back-end for a TES endpoint. The current setup is tuned for a backend that looks and acts like a traditional HPC cluster with a shared filesystem. It was also  communicated
on the calls that the intended usage was "stage data -> work -> checkpoint data"  but the structure allows for "stage data -> work -> work -> work -> work -> work -> checkpoint data -> work -> checkpoint data" Not a great design. 

Are are some perfectly good alternates: 

1 - for HPC backend you have a shared FS, data staging taken care of already by volume mounts
2 - for individual serial workflows, just package the entire pipeline in a single container and make the script run each tool serially
3 - for a TES backend that works in the way that AWS Batch works, launch a bootstrap container that stages data as necessary and launches a sibling container to do the work.

buchanae:
Thanks @delagoya 

>  "stage data -> work -> work -> work -> work -> work -> checkpoint data -> work -> checkpoint data"

As I understand "checkpoint data" means "upload files to storage", but in Funnel at least, uploading only happens once at the end of the task. What am I misunderstanding?

buchanae:
> 2 - for individual serial workflows, just package the entire pipeline in a single container and make the script run each tool serially

[...]

PR #95, opened 2017-10-24
Clarify volumes docs.

buchanae:
#64 and #79 both comment on understandable confusion around the volumes field. This attempts to add more docs around it, but maybe it's still not enough. This is a confusing feature that seems to arise from needing to stitch together multiple container filesystems.

Would love feedback from @mbookman and @erikvdbergh on this one.

### DIFF
--- a/task_execution.proto
+++ b/task_execution.proto
@@ -53,9 +53,17 @@ message Task {
 
   // OPTIONAL
   //
-  // Declared volumes.
-  // Volumes are shared between executors. Volumes for inputs and outputs are 
-  // inferred and should not be declared here.
+  // Volumes are directories which may be used to share data between
+  // Executors. Volumes are initialized as empty directories by the
+  // system when the task starts and are mounted at the same path
+  // in each Executor.
+  //
+  // For example, given a volume defined at "/vol/A",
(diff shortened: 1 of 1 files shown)


mbookman:
Thank-you, Alex. The comment added makes the usage clear to me.

There is something about the name `volumes` and the actual usage that seems unexpected, though it is hard to put my finger on it. I think with "volume" I am expecting either specification of a disk (with size and disk type) or specification of both ends of the -v (host and docker paths).

Thinking of some other naming suggestions, `disk_paths` or `directories` might be a bit simpler. That the description you have added of a `volumes` is that they are `directories` points me to that being a better name. Would there be any confusion if they were simply called `directories`?

pditommaso:
👍

buchanae:
@mbookman Ya, the term `volumes` is overloaded. We'll try to come up with something more straightforward. Thanks for the suggestions.

Other ideas:
- `shared_directories`
- `common_directories`
- `common_data`

Also, we could define one single path that implementations should share. Also, there's still the discussion of #59 which would affect this.

buchanae:
Sorry, I think this was merged at the wrong time. Luckily it's just documentation. I've created an issue for renaming volumes in #100

PR #105, opened 2018-06-05
use /ga4gh/tes service prefix

### LINKED ISSUE #68: API endpoint base URL
In https://github.com/ga4gh/workflow-execution-schemas/pull/9#issuecomment-285988194, an approach to API endpoint URL naming was mentioned. We should discuss and agree on adopting this pattern for TES.

The pattern would prefix service API endpoints with `/ga4gh/tes/v1` e.g. the cancel endpoint would be `/ga4gh/tes/v1/tasks/1234:cancel`

buchanae:
closes #68

### DIFF
--- a/task_execution.proto
+++ b/task_execution.proto
@@ -536,14 +536,14 @@ service TaskService {
   // other documentation.
   rpc GetServiceInfo(ServiceInfoRequest) returns (ServiceInfo) {
     option (google.api.http) = {
-      get: "/v1/tasks/service-info"
+      get: "/ga4gh/tes/v1/service-info"
     };
   }
 
   // Create a new task.
   rpc CreateTask(Task) returns (CreateTaskResponse) {
     option (google.api.http) = {
-      post: "/v1/tasks"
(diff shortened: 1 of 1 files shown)


buchanae:
Note that I moved the service-info endpoint here. Now that we have a service namespace, this doesn't need to fall under that "tasks" namespace.

aniewielska:
Generally, it is nice that WES and TES now have a common URL pattern, plus `serviceInfo` not under tasks is cleaner. But is there a way of actually extracting a base path, as in WES: ```basePath: /ga4gh/wes/v1```?

delagoya:
Yes, we should strive for consistency across Cloud WS APIs. There is a general schema cleanup coming that will do a couple of things:

1. Make the swagger definition the canonical spec
2. Remove the protobuf schema and makefile
3. clean up the swagger schema to remove a lot of the protobuf artifacts (mainly `title`)

After those are pulled in, we will make structural changes like the above. 

Move to **close** this PR in favor of new issues and PRs that focus on openapi schema.

delagoya:
This was solved in #113

PR #107, opened 2018-06-05
remove paused state

### LINKED ISSUE #99: Proposal: remove paused state
Discussion continued from #89

buchanae:
closes #99 

We've flip-flopped on this a bit, but I'm in favor of removing things from the spec that no one is currently using and no one currently has a plan or specification for how to use.

If someone has a robust plan for how tasks can be paused, we can add this back.

### DIFF
--- a/task_execution.proto
+++ b/task_execution.proto
@@ -372,11 +372,6 @@ enum State {
   // has been started.
   RUNNING = 3;
 
-  // The task is paused.
-  //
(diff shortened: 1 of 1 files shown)


susheel:
@buchanae Could you point this PR to the new YAML spec in `develop`

vsmalladi:
We reopen if there is new discussion

PR #125, opened 2019-06-18
Fix link and file name for swagger definition

osallou:


aniewielska:
Thanks for the contribution. In the meantime, however, the OpenAPI version has been updated and the new spec file is now (hopefully) properly referenced in READ.me.

PR #132, opened 2020-08-06
OpenAPI 3

aniewielska:
The result of conversion from OpenAPI 2 to 3 using https://converter.swagger.io/ without any changes
Checks so far: 
- https://validator.swagger.io/validator/?url=https%3A%2F%2Fraw.githubusercontent.com%2Faniewielska%2Ftask-execution-schemas%2Fopenapi3%2Fopenapi%2Ftask_execution_service.openapi.yaml
It seems to pass validation (unlike WES).
- I ran the new spec through OpenAPI codegen for Java and got roughly the same wire format (minus ServiceInfo which now has lots of new fields). Also required seems now to be acknowledged by the generator for Java 
TODOs:
- compare with WES and try to align
I noticed that WES additionally has:
-- Error code responses listed (400, 404 etc) - not sure, if we want to add it at that stage. Breaking change.
-- Error response model - not sure, if we want to add it at that stage. Breaking change
-- URL defined as composite of host, path This we can do for TES.
-- Executive Summary - this is what we need for docs generation
- ~~externalise ServiceInfo~~ - DONE. I added additional storage and moved '/service-info' outside of '/tasks'. That is breaking change as well, but seems required.

### DIFF
--- a/openapi/task_execution_service.openapi.yaml
+++ b/openapi/task_execution_service.openapi.yaml
@@ -1,0 +1,575 @@
+openapi: 3.0.1
+info:
+  title: Task Execution Service
+  version: 0.4.0
+servers:
+- url: /ga4gh/tes/v1
+paths:
+  /tasks:
+    get:
+      tags:
+      - TaskService
+      summary: |-
+        List tasks.
+        TaskView is requested as such: "v1/tasks?view=BASIC"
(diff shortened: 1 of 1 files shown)


aniewielska:
Version update done.

PR #133, opened 2020-08-07
New docs build system

aniewielska:
This is a preview of how to add ReDoc to TES.
(1) It builds on top of OpenAPI3, so it contains changes from #132 and builds on top of them (it would be good to accept the Open API 3 spec first and later add docs).
(2) Also the build system assumes the existence of the `gh-pages` branch, so that needs to be created beforehand.
(3) The actual needed changes are:
- .spec-docs.json
- .travis.yml
 
(4) I also changed the spec a bit to improve the layout of generated docs, but still the result looks differently from the WES one. I am not sure, if the ga4gh theme worked or if it is expected to work for a fork.
 
The result is not any pretty yet and can be seen here: https://aniewielska.github.io/task-execution-schemas/preview/new-docs-build/docs/

### DIFF
--- a/.spec-docs.json
+++ b/.spec-docs.json
@@ -1,0 +1,4 @@
+{
+  "apiSpecPath": "openapi/task_execution_service.openapi.yaml",
+  "redocTheme": "ga4gh"
+}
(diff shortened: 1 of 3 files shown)


aniewielska:
I pushed the code directly to a branch in this repository and will create a new PR from branch instead.

PR #134, opened 2020-09-15
New docs build system

aniewielska:
This PR brings machinery of building ReDoc to TES.
I created an empty orphaned `gh-pages` branch and set up GITHUB_TOKEN in Travis.
As a result the docs are being generated, ~~but still the result is visually different than the one for WES.~~ and finally they resemble the ones for WES! (The package version mentioned in the migration instructions was wrong).
You can see the results here: http://ga4gh.github.io/task-execution-schemas/preview/new-docs-build/docs/
The missing parts - reorganising parts of description/titles and adding more executive summary.
Also it would be nice to add links to the docs to READ.me

PR #135, opened 2020-09-15
OpenAPI Doc content improvements

kellrott:


aniewielska:
Looks like we are working on the same topic. Could you possibly please have a look at #134 first and perhaps it is time to approve it? I kept `summary` items short (such as GetTask) similarly to what WES had, but I left descriptions largely intact. It should help with your work here.

kellrott:
I've sync'd with origin/master after merging #134. 

It looks like the operation `summary` field is used to populate the overview list on the left hand side of the doc, so those should probably be kept to a single word.

kellrott:
I've added a `Introduction` and `Standards` section to the head of the document that are a new copy/paste of the blocks found in the WES document.

aniewielska (approved):
LGTM (and to the OpenAPI validator, too)

PR #138, opened 2020-10-14
Client implementations

aniewielska:
And we mention also the client implementations of TES. For workflow managers I linked TES backends.


kellrott (approved):
LGTM

PR #143, opened 2020-11-23
Introduce fixed service info type

aniewielska:
This PR sets the value of ServiceType artifact to `tes`. Looks decently in ReDoc and in OpenAPI generator, but I am not sure, if this is the best possible option in OpenAPI.
Work inspired by [htsget](https://github.com/samtools/hts-specs/pull/493/files), as there seems to be no good examples in Cloud Workstream.

### DIFF
--- a/openapi/task_execution_service.openapi.yaml
+++ b/openapi/task_execution_service.openapi.yaml
@@ -518,6 +518,17 @@ components:
             type: string
           example: us-west-1
       description: Resources describes the resources requested by a task.
+    tesServiceType:
+      allOf:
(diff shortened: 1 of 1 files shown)


kellrott:
Are we trying to get this in before 1.0?

aniewielska:
It should not change the payload of the spec, other than by making `tes` the only option for the service type (artifact). But I would like to get some attention of TASC, as the solution does not look particularly elegant (> 13 lines of code just to say, that only one option is possible for one field) and would need to be repeated in all other APIs implementing Service Info.

jb-adams (approved):
looks good to me, essentially the same as it appears in htsget. I agree the solution doesn't look particularly elegant, but it seems to be the only way to do it in OpenAPI.

aniewielska:
I will merge it now then, as it essentially does not change the wire format (implementations were supposed to be returning `tes` anyway).

PR #144, opened 2020-11-23
Deduplicate view param

aniewielska:
This PR externalises the `view` parameter of list and get task. ~There is some problem generating GH pages after this change in my fork (but the files got generated in gh-pages), so let's see, if the problem is universal~ (it worked).


kellrott (approved):
Seems to render docs correctly.

aniewielska:
@kellrott - please check again. I had to rebase the change on top of others and it looks validating correctly, but would like someone else to have a look as well. Also I reapplied your changes of field names in the comment of view parameter, hope no errors there as well. If it look OK to you, please merge it.

PR #147, opened 2020-11-24
The lost logo

aniewielska:
I replaced the missing logo with the current one from GA4GH website, but not sure, if this is the best solution. It fixes the problem, but the logo may well disappear again.


kellrott:
LGTM

kellrott (approved):

rishidev:
GA4GH Comms has suggested using https://ga4gh.org/wp-content/uploads/GA-logo-horizontal-RGB-1400.png
The other files are from the template theme. If the SVG is preferred that could also be uploaded instead I guess.

rishidev:
The permalink for the logo has now been confirmed as https://w3id.org/ga4gh/ga4gh-logo.svg

aniewielska:
At the moment the old logo is working fine, but I replaced the link used in this PR with the permalink anyway.

rishidev:
Thanks, this will future proof this.

PR #148, opened 2020-11-24
Mention DRS, TRS

aniewielska:
Following the pattern of WES (slightly dated), I added Possible Future Enhancements with DRS and TES integration. Might be a good place to mention other items from the Roadmap or link a roadmap?


kellrott (approved):
LGTM

PR #149, opened 2020-12-09
Security disclaimer for implementations

aniewielska:
As requested by REWS, copied from @rishidev email (no bold, though). Please check that we want it styled like that (was that italics intentional?)


rishidev:
Thanks for the changes. The italics was part of the text, but IMHO are optional, so fine either way.

rishidev (approved):
LGTM

PR #152, opened 2021-02-09
Bump version to 1.0.0

aniewielska:
This is one of the elements missing post approval. Not sure, if there is anything else in the repository that we should change in addition to the version in the spec?
This PR would be the last one towards `master` and after that we will update develop and make it the default branch with further PRs targeting develop and not master.
After approving this PR and sanitising the branches, it would also be nice to prepare a release. 
Further adoption of GitFlow/HubFlow can be discussed later. Essentially, it introduces release branches, but only just about when a further release is planned and potentially feature branches for collaborative work. For PRs from forks (which we now state are our preference) a default develop branch will be sufficient.


kellrott:
Should we leave the `task_execution.swagger.yaml` file? It's still at 0.4.0. Maybe it's time to remove it for the 1.0.0 release.

aniewielska:
I was about to ask about it and forgot. I vote for removing it, although for some reason WES decided to keep OA2 version, which I believe diverged from OA3. I will delete it as of this PR.

kellrott:
:+1:

aniewielska:
@kellrott - if it looks OK to you, please merge the PR.

kellrott (approved):
LGTM

PR #154, opened 2021-06-28
Proposal: add "backend_parameters" to tesResources

MattMcL4475:
According to the WDL spec, "The runtime section defines key/value pairs for runtime information needed for this task":

https://github.com/openwdl/wdl/blob/master/versions/1.0/SPEC.md#runtime-section

This commit extends tesResources to support an arbitrary dictionary of additional properties, which ensures WDL parsers can pass-through their full list of runtime key/value pairs to a TES backend. The canonical example of its use would be for WDL authors to be able to specify a specific cloud VM SKU in their WDL files.

Please see previous closed PR for discussion: https://github.com/ga4gh/task-execution-schemas/pull/127


aniewielska (approved):
We discussed it thoroughly and this version looks good to me.

PR #155, opened 2021-09-02
Making file type an optional argument

kellrott:


openapi/task_execution_service.openapi.yaml
```diff
@@ -370,14 +370,17 @@ components:
       description: ExecutorLog describes logging information related to an Executor.
     tesFileType:
       type: string
+      description: |-
+        Define if input/output element is a file or a directory. It is not required
+        that the user provide this value, but it is required that the server fill in the
+        value once the information is avalible at run time.
```
m4nu56, line 376: small typo with "avalible"

aniewielska:
If input/output types are not mandatory, should they have a default value then? How does a default behave for an optional parameter? 
I will check a code generator for this, as I expect problems.

aniewielska (commented):
I am not sure of the behaviour of the default values for optional parameters, especially in the context of generated code. I will check it with code generation.

uniqueg:
Any feedback on your tests, @aniewielska? Intuitively I would think that handling defaults should be pretty standard for any validator and code generator based on JSON Schema...

aniewielska:
I think at the moment the type is still mandatory, if we leave defaults with optional. I think the spec was wrong previously in combining defaults and required. But the behaviour we want is no default. When no type of in/output is provided the implementation needs to figure it out itself and not assume it is FILE.

uniqueg:
Given that the accepted values are defined in an enum, the value passed for  `tesFileType` _has_ to be a value matching one of the enumerated values. If there's only `FILE` and `DIRECTORY` in the enum, one couldn't send a valid request omitting this if there's no default set. I would propose adding `UNSET` or `UNSPECIFIED` or something to that effect to the enum and then use that as default.

aniewielska:
Right. So, if we cannot have null enums, I vote for adding `UNSPECIFIED` and making it default as it nicely described th intention here that the caller would not like to specify the type.

mr-c (approved):

pditommaso (approved):

Issue #156, opened 2021-09-02
Indicate Streamable files


kellrott:
Provide a flag that indicates that a file could be mounted via a streaming interface, if it is available, without causing performance degradation to the code being deployed.

uniqueg:
Requested by Greek ELIXIR node, see https://docs.google.com/spreadsheets/d/1vBFhBQ-nFqhSL5dLjQfOWO6x9BzmV9x6l18p9GYRZdQ/edit#gid=0

Contacts: @zagganas & @vergoulis

vsmalladi:
Closed by #157

PR #157, opened 2021-09-02
Adding streamable flag to tesInput

kellrott:
#156


openapi/task_execution_service.openapi.yaml
```diff
@@ -414,6 +414,17 @@ components:
...
             If content is not empty, "url" must be ignored.
+        streamable:
+          type: boolean
+          description: |-
+            Indicate that a file resource could be accessed using a streaming
+            interface, ie a FUSE mounted s3 object. This flag indicates that
+            using a streaming mount, as opposed to downloading the whole file to
+            the local scratch space, may be faster despite the latency and
+            overhead. This does not mean that the backend will use a streaming
+            interface, as it may not be provided by the vendor, but it the
```
m4nu56, line 425: small typo: "...but iF the capacity is avaIlable..."

aniewielska (commented):
Does streamable make sense for inputs only or outputs as well?

vsmalladi:
@aniewielska I think inputs only makes sense. Since the goal is to either download the files or not. Outputs need to be published so they can be captured later.

mr-c (approved):

PR #176, opened 2022-06-27
added callback specs

sgalpha01:
I've added callback schema according OpenAPI v3 specs. Kindly review it and suggest necessary changes, if required.


openapi/task_execution_service.openapi.yaml
```diff
@@ -146,6 +146,21 @@ paths:
...
+      callbacks:
+        statusChange:
+          '{$request.body#/callback_url}':
+            post:
+              requestBody:
+                description: ""
+                content:
+                  application/json:
+                    schema:
+                      $ref: '#/components/schemas/tesTaskStatusChangeNotification'
```
uniqueg, line 158: That's quite a mouthful. Perhaps better: `tesCallbackStatus`?

openapi/task_execution_service.openapi.yaml
```diff
@@ -654,6 +669,12 @@ components:
           type: string
           description: |-
             Optional user-provided description of task for documentation purposes.
+        callback_url:
+          type: string
+          format: uri
+          description: |-
+            Optional user-provided callback URL for task status updates.
```
uniqueg, line 676: Note that this should probably _not_ be shared together with other info when users call, e.g., `GET /tasks/{task_id}`. As there are, unfortunately, no separate models/schemas for `POST`ing a task request and `GET`ting info associated with a task, I'm afraid you can only document that in the description for now (and probably should).

uniqueg: Also, perhaps it is unnecessary, here, to be explicit about what the `callback_url` can be used for. Possibly there will be other uses for it in the future.

openapi/task_execution_service.openapi.yaml
```diff
@@ -146,6 +146,21 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/tesCreateTaskResponse'
+      callbacks:
+        statusChange:
+          '{$request.body#/callback_url}':
+            post:
+              requestBody:
+                description: ""
```
uniqueg, line 154: Please add a description.

uniqueg (changes requested):
Looks good but some points need some work, see comments.

openapi/task_execution_service.openapi.yaml
sgalpha01: Added the description

openapi/task_execution_service.openapi.yaml
```diff
@@ -654,6 +671,14 @@ components:
           type: string
           description: |-
             Optional user-provided description of task for documentation purposes.
+        callback_url:
+          type: string
+          format: uri
+          description: |-
+            Optional user-provided callback URL. Note that the URL should be
+            provided while creating a new task, and it won't be displayed when
```
uniqueg, line 679: Two minor wording issues:
> Note that the URL _can_ be provided (instead of: Note that the URL _should_ be provided)

> _but it MUST NOT_ be displayed (instead of: _and it won't_ be displayed)

uniqueg (changes requested):

kellrott:
I'm not sure if this PR should be targeting `develop`. That is the current destination for TES 1.1. I don't think that callbacks will be part of 1.1. Maybe we should create a `develop-1.2` branch to hold these more forward looking features.

openapi/task_execution_service.openapi.yaml
```diff
@@ -146,6 +146,23 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/tesCreateTaskResponse'
+      callbacks:
+        statusChange:
+          '{$request.body#/callback_url}':
+            post:
+              description: |-
+                The API will send notifications to the callback URL, if provided,
```
aniewielska, line 154: API is an interface. I suggest changing it to `TES server`

openapi/task_execution_service.openapi.yaml
```diff
@@ -654,6 +671,14 @@ components:
           type: string
           description: |-
             Optional user-provided description of task for documentation purposes.
+        callback_url:
+          type: string
+          format: uri
+          description: |-
+            Optional user-provided callback URL. Note that the URL can be
+            provided while creating a new task, but it MUST NOT be displayed when
```
aniewielska, line 679: I think we should return the callback URL together with the rest of task payload. Other information contained in the task details can be equally sensitive. The only element to think about is whether it should be returned only in FULL or already in BASIC (probably BASIC makes more sense).

openapi/task_execution_service.openapi.yaml
```diff
@@ -741,6 +766,19 @@ components:
           example: 2020-10-02T10:00:00-05:00
           readOnly: true
       description: Task describes an instance of a task.
+    tesCallbackStatus:
+      type: object
+      required:
+      - task_id
+      - state
+      properties:
+        task_id:
```
aniewielska, line 775: For consistency we should call this parameter (`task_id`) just `id` as it is called elsewhere.

aniewielska (changes requested):

aniewielska:
BTW, we are quite inconsistent in the choice of camel vs snake case, but it has happened before this PR.

openapi/task_execution_service.openapi.yaml
```diff
@@ -741,6 +764,19 @@ components:
           example: 2020-10-02T10:00:00-05:00
           readOnly: true
       description: Task describes an instance of a task.
+    tesCallbackStatus:
```
uniqueg, line 767: Looks to me like the `schemas` are listed in alphabetical order, so the `tesCallbackStatus` should be inserted as the first schema, before `tesCancelTaskResponse` instead.

uniqueg (changes requested):

openapi/task_execution_service.openapi.yaml
```diff
@@ -741,6 +764,19 @@ components:
...
+      type: object
+      required:
+      - task_id
+      - state
+      properties:
+        id:
```
uniqueg, line 773: Inconsistent with `task_id` mentioned just above among the `required` properties. Change here or above.

uniqueg:
@aniewielska: You may be right about `snake_case`/`camelCase` inconsistency, but I don't see any new inconsistencies being introduced here. The PR follows the examples in the [OpenAPI 3.0 docs](https://swagger.io/docs/specification/callbacks/) - which uses `snake_case` for schema properties and `camelCase` for pretty much all else.

openapi/task_execution_service.openapi.yaml
sgalpha01: Yup, sorry I missed it :)

openapi/task_execution_service.openapi.yaml
```diff
@@ -146,6 +146,23 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/tesCreateTaskResponse'
+      callbacks:
+        statusChange:
+          '{$request.body#/callback_url}':
```
coverbeck, line 151: This cool feature was mentioned at the GA4GH call, so I took a peek... I'm curious, how is authentication supposed to work for the callback url?

aniewielska: The idea here was not to put callback url behind extra authentication (the URL itself might be a secret?) and additionally advise consumers of the callback to verify the information by hitting the GET endpoint.

patmagee: @aniewielska thanks for the clarification. I wonder if it would still potentially make sense to make this an object to future proof in the case of wanting to add additionally access mechanisms in the future

coverbeck: > The idea here was not to put callback url behind extra authentication (the URL itself might be a secret?) and additionally advise consumers of the callback to verify the information by hitting the GET endpoint.

I think it would be nice to have your clarification in the spec. 

I'm still slightly concerned about needing to expose a public callback url without auth, and I think Patrick's proposal addresses that, but I don't feel too strongly (and I'm not a reviewer anyway, just lurking :) ).

openapi/task_execution_service.openapi.yaml
```diff
@@ -219,6 +236,19 @@ components:
         - FULL
 
   schemas:
+    tesCallbackStatus:
```
coverbeck, line 239: How will the the callback know which TES server is making the request? Is the callback expected to look at the Origin header (but I don't think that's guaranteed to be set)?

openapi/task_execution_service.openapi.yaml
uniqueg: Good points here, thanks. Agree with future proofing by making it an object and extending the description to clarify that clients should verify the info from the callback and that the callback URL may include a secret

PR #185, opened 2022-08-11
feat: allow wildcards in output paths

### LINKED ISSUE #77: Proposal: add a GLOB value to FileType enum
The purpose of this proposal is the extend the ability of TES API to handle dynamically generated output files which name may not be known at the time the task definition by adding a `GLOB` value to the possible types of a `TaskParameter` output. 

This would allow the possibility to specify as output one or more files matching a [glob pattern](http://tldp.org/LDP/abs/html/globbingref.html) eg. `*.bam` or `sample_x.{bam,bai}`, etc. 

The server would resolve the list of matching files and would return them to the client when a `FULL` task view is retrieved. 

The output `name` field can be used to logically group the files that resolves against the same glob pattern. For example having the following output in the task definition: 

```
output[ {name: "bam", path: "*.bam", type: "GLOB" } ]
```

the task view would return: 

```
output[ 
   {name: "bam", path: "/some/path/pair_x.bam", type: "FILE" },
   {name: "bam", path: "/some/path/pair_y.bam", type: "FILE" },
   {name: "bam", path: "/some/path/pair_z.bam", type: "FILE" }
 ]  
```

uniqueg:
Fixes #77

### Description

This PR adds provisions that allow clients to specify pathname matching wildcards ("globs") when specifying task outputs.

It addresses the discussion points summarized in [this comment](https://github.com/ga4gh/task-execution-schemas/issues/77#issuecomment-1198253120) in the following ways:

- **Should TES implementations be required to support pathname matching in task outputs?**
   Yes. I feel that the use case is sufficiently common that it justifies the added complexity incurred by providing native support in TES, especially since the lack of support for globbing has been a blocker for a more widespread adoption of TES for some upstream implementers, e.g., Nextflow. This proposal adds the provisions for such native support.

- **How to indicate that TES should interpret a specified output as a glob?**
   This proposal requires TES implementations to always apply pathname matching rules to `tesOutput.path` values, unless wildcards are explicitly escaped. This behavior is familiar from (most) shells, completely removes any ambiguities and only alters the `tesOutput.path` description, so no structural changes to the specification are necessary. However, depending on a TES implementer's interpretation of the current specification, it may constitute a breaking change. Specifically, `tesOutput.path` values containing pathname matching wildcards would be treated differently by a TES implementation adopting the proposed change if (and only if) it previously interpreted wildcards literally. However, the expected behavior for paths including wildcards is currently underspecified (implementations may choose to expand or interpret them literally), so the proposed change rather adds clarity to a potential source of ambiguity across implementations.

- **Which globbing rules to prescribe?**
   This proposal suggests prescribing POSIX/Open Group pathname/pattern matching rules, because they are formally specified (see [here](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_13) and [here](https://pubs.opengroup.org/onlinepubs/007908799/xsh/glob.html)), making them compatible with any POSIX-compatible shell environment, including Bash.

- **How to constuct remote storage URLs?**
   This proposal requires clients to do the following, whenever wildcards are used in `tesOutput.path`:

  1. specify a directory for `tesOuput.url`
  2. provide a `tesOutput.path_prefix` (new property) to indicate to TES implementations a prefix to remove from matched pathnames in order to identify the subdirectory tree that is then recreated at the directory specified for `tesOuput.url`  

   This solution fully supports wildcards in any part of `tesOutput.path` while avoiding name clashes, maximizing the control clients have over the eventual location of outputs and being straightforward to implement on the server side.

### Examples

* Use wildcards to copy files from multiple directories whose names are not known (new behavior).

   Outputs to be copied:

   ```console
    /path/in/container/unknown_subdir_1/output.bam
    /path/in/container/unknown_subdir_1/output.sam
    /path/in/container/unknown_subdir_2/output.bam
    /path/in/container/unknown_subdir_2/output.sam
   ```

   Schema values to supply:

   ```yaml
   tesOutput.path: /path/in/container/*/*.?am
   tesOutput.path_prefix: /path/in/container
   tesOutput.url: https://my.storage.org/bucket/
   ```

   URLs of outputs:

   ```console
   https://my.storage.org/bucket/unknown_subdir_1/output.bam
   https://my.storage.org/bucket/unknown_subdir_1/output.sam
   https://my.storage.org/bucket/unknown_subdir_2/output.bam
   https://my.storage.org/bucket/unknown_subdir_2/output.sam
   ```

* Copy a file whose name contains wildcard characters (behavior previously unspecified).

   Output to be copied:

   ```console
   /path/with/wildc*rd_char?
   ```

   Schema values to supply:

   ```yaml
   tesOutput.path: /path/with/wildc\*rd_char\?
   tesOutput.path_prefix: # ignored
   tesOutput.url: https://my.storage.org/bucket/wildc*ard_char?
   ```

   URL of output:

   ```console
   https://my.storage.org/bucket/wildc*ard_char?
   ```
  
* Copy output directory whose name is known (behavior unchanged).

   Output to be copied:

   ```console
   /path/on/container/output/
   ```

   Schema values to supply:

   ```yaml
   tesOutput.path: /path/on/container/output/
   tesOutput.path_prefix: # ignored
   tesOutput.url: https://my.storage.org/bucket/output/
   ```

   URL of output:

   ```console
   https://my.storage.org/bucket/output/
   ```

### DIFF
--- a/openapi/task_execution_service.openapi.yaml
+++ b/openapi/task_execution_service.openapi.yaml
@@ -450,7 +450,9 @@ components:
         url:
           type: string
           description: |-
-            URL for the file to be copied by the TES server after the task is complete.
+            URL at which the TES server makes the output accessible after the task is complete.
+            When tesOutput.path contains wildcards, it must be a directory; see
+            `tesOutput.path_prefix` for details on how output URLs are constructed in this case.
             For Example:
              - `s3://my-object-store/file1`
              - `gs://my-bucket/file2`
@@ -458,8 +460,18 @@ components:
         path:
           type: string
           description: |-
-            Path of the file inside the container.
-            Must be an absolute path.
+            Absolute path of the file inside the container.
+            May contain pattern matching wildcards to select multiple outputs at once, but mind
+            implications for `tesOutput.url` and `tesOutput.path_prefix`.
+            Only wildcards defined in IEEE Std 1003.1-2017 (POSIX), 12.3 are supported; see
+            https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_13
+        path_prefix:
+          type: string
+          description: |-
+            Prefix to be removed from matching outputs if `tesOutput.path` contains wildcards;
+            output URLs are constructed by appending pruned paths to the directory specfied
+            in `tesOutput.url`.
+            Required if `tesOutput.path` contains wildcards, ignored otherwise.
         type:
           $ref: '#/components/schemas/tesFileType'
       description: Output describes Task output files.


uniqueg:
Would appreciate feedback @kellrott @aniewielska @pditommaso @mr-c @MattMcL4475 @vsmalladi @wleepang @geoffjentry 🙏🏻

mr-c (approved):

mr-c:
Looks good to me; tagging @tetron for his Arvados perspective and additional CWL perspective

wleepang:
I think this looks good, but I'm curious to see some concrete examples. For instance, if I set:

```
tesOutput.path: /path/to/folder
tesOutput.path_prefix: /path/to/folder
yes output.url: s3://bucketname/my/results
```

Should I expect the contents of `/path/to/folder` to be copied recursively to `s3://bucketname/my/results`?

vsmalladi:
@wleepang If i understand this correctly this is what I would expect

tesOutput.path: /path/to/folder/data
tesOutput.path_prefix: /path/to/folder
output.url: s3://bucketname/my/results

URL: s3://bucketname/my/results/data

@uniqueg can you confirm this is what is expected with the combination of `output_url` and `tesOutput.path_prefix`?

uniqueg:
@wleepang @vsmalladi 

I have now added two examples to the PR description (my bad for not doing so earlier!).

Note that in the examples you give, the path doesn't include any wildcards, so `tesOutput.path_prefix` should be ignored by the TES implementation. The URLs would be treated according to the current rules, which means that TES would interpret the value of `tesOutput.url` as a fully qualified URL pointing to the object to output. So in your example, @wleepang, contents of `/path/to/folder` would indeed be recursively copied to `s3://bucketname/my/results`. In your case though, @vsmalladi, contents of `/path/to/folder/data` would be copied to `s3://bucketname/my/results`, *not* `s3://bucketname/my/results/data`.

Or at least this is my interpretation of the current specs.

vsmalladi:
@uniqueg Thanks for the clarification.

Then LGTM.

mr-c (approved):

PR #189, opened 2022-09-30
Add CANCELLATION_REQUESTED state to delete cloud resources

MattMcL4475:
Currently TES has a `CANCELED` state.  However, cancelling a task means the associated cloud resources need to be deleted, which can take a few minutes to complete - much longer than is reasonable for an HTTP request timeout.

Proposal: Add a new TES state: `CANCELLATION_REQUESTED`.  When a caller calls the `CancelTask` operation, TES shall return `CANCELED` if and only if all dependent/downstream resources have been deleted.  If it cannot delete the downstream cloud resources during the HTTP request lifetime, it shall return `CANCELLATION_REQUESTED`, until all downstream resources have been deleted.

Example: in Cromwell on Azure, a user can submit a workflow that runs one task.  If a user cancels that workflow while the task is still running, currently, Cromwell will call "CancelTask", and TES will return `CANCELED`. However, at this point, the Azure Batch job and pool have NOT been deleted yet, since this is done by a background thread running on the TES web server.  This is problematic because Cromwell cannot observe whether TES' expensive resources have been deleted yet, which can leave the system in a broken+expensive state, with VMs sitting idle in Azure Batch.  A control plane that is managing the Cromwell environment (such as Terra) can also therefore not observe whether the system has successfully completely cancelled all in-flight workflows.

### DIFF
--- a/openapi/task_execution_service.openapi.yaml
+++ b/openapi/task_execution_service.openapi.yaml
@@ -622,7 +622,9 @@ components:
          - `SYSTEM_ERROR`: The task was stopped due to a system error, but not from an Executor,
         for example an upload failed due to network issues, the worker's ran out
         of disk space, etc.
-         - `CANCELED`: The task was canceled by the user.
+         - `CANCELED`: The task was canceled by the user, and downstream resources have been deleted.
+         - `CANCELLATION_REQUESTED`: The task was canceled by the user,
+        but the downstream resources are still awaiting deletion
       default: UNKNOWN
       example: COMPLETE
       enum:
@@ -635,6 +637,7 @@ components:
       - EXECUTOR_ERROR
       - SYSTEM_ERROR
       - CANCELED
+      - CANCELLATION_REQUESTED
     tesTask:
       required:
       - executors


uniqueg (changes requested):
Great suggestion! For consistency with WES, I would prefer to name the state `CANCELING` though. See [enum definition](https://github.com/ga4gh/workflow-execution-service-schemas/blob/c3b19854240c4fcbaf3483e22b19db0a918a7ee5/openapi/components/schemas/State.yaml) and [PR](https://github.com/ga4gh/workflow-execution-service-schemas/pull/104) in WES.

@kellrott: seems that the CI only works for PRs that do not originate from forks. I'll see if I find some time to fix that in the coming days (and by fix, I probably mean disabling the failing actions for PRs from forks)

kellrott:
Can you re-target the PR to develop-1.1? I'm trying to collect all changes into that, so I can run on a final merged PR with CI before merging into develop.

MattMcL4475:
Thank you @uniqueg and @kellrott for the quick feedback!  Changed it to `CANCELING` and re-targeted `develop-1.1`

uniqueg (approved):
Looks good to me. Maybe you could add a full stop to the description of the new state, as it seems the other state descriptions have one (at least the ones I could see in the diff)

MattMcL4475:
Thanks @uniqueg ; @kellrott looks like it's ready to merge

Issue #194, opened 2023-08-16
Clarification required: tesOutput.path_prefix


vsmalladi:
In the spec it appears that `tesOutput.path_prefix` is ignored if `tesOutput.path` doesn't contain wildcards.

However, when there are no wildcards provided we are still pruning the paths of the parent to just include the child.

```
tesOutput.path: /path/on/container/output/
tesOutput.path_prefix: # ignored
tesOutput.url: https://my.storage.org/bucket/output/

prune:  /path/on/container/
```

This selective use of `path_prefix` seems confusing and what we be more consistnt to implement is always use the `path_prefix` such that you could do this:

```
tesOutput.path: /path/on/container/output/
tesOutput.path_prefix: /path/on
tesOutput.url: https://my.storage.org/bucket/container/output/
```

Thoughts @uniqueg, @kellrott , @wleepang ?

MattMcL4475:
A few more clarifications to discuss:

1.  It `type` is not specified for a `tesOutput`, do we assume it defaults to "FILE" ?

2.  File specified with trailing slash:

```
tesOutput.tesFileType = "FILE"
tesOutput.path: /path/on/container/output/
```

Ideally: TES should return HTTP status code 400 "Bad Request"
Alternatively: TES task state should be set to "EXECUTOR_ERROR"
Reasoning: a file can never end in a trailing slash

3.  tesFileType of "FILE" is misleading when it's a pattern:

```
tesOutput.tesFileType = "FILE"
tesOutput.path: /path/on/container/output/*.bam
```

It seems we should add a tesFileType of "PATTERN" for this use case.  "FILE" doesn't make sense, because it's really a pattern, and it results in an array of files. 

4.  "DRY" - two ways of doing the same thing, correct?  Maybe this is fine but would like to confirm they are equivalent.
Are these equivalent?
```
tesOutput.tesFileType = "FILE"
tesOutput.path: /path/on/container/output/*
```

```
tesOutput.tesFileType = "DIRECTORY"
tesOutput.path: /path/on/container/output/
```

5.  Does TES consider these as the same or different?  For example, in RSYNC, the first example would include "output/" in the file names, while the second example would not.

```
tesOutput.tesFileType = "DIRECTORY"
tesOutput.path: /path/on/container/output
```

```
tesOutput.tesFileType = "DIRECTORY"
tesOutput.path: /path/on/container/output/
```

[...]

PR #195, opened 2023-08-17
Remove path_prefix from TES specification

MattMcL4475:
`path_prefix` as currently defined in the specification is ambiguous.  It is also a convenience feature, since the local filesystem can be reorganized as needed either by including additional bash commands in the task `command` or by including additional executors.  Therefore, we propose removing it from the specification, as the result of a conversation between @giventocode @vsmalladi @uniqueg @kellrott on 8/17/2023

Addresses: https://github.com/ga4gh/task-execution-schemas/issues/194

### DIFF
--- a/openapi/task_execution_service.openapi.yaml
+++ b/openapi/task_execution_service.openapi.yaml
@@ -525,8 +525,7 @@ components:
           type: string
           description: |-
             URL at which the TES server makes the output accessible after the task is complete.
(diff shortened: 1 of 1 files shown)


BMurri:
It appears that ga4gh/task-execution-schemas#194 isn't [entirely settled](https://github.com/ga4gh/task-execution-schemas/issues/194#issuecomment-1686643353). This should be deferred until the final resolution is determined.

PR #199, opened 2023-10-04
auth: mandate JWT bearer token auth

### LINKED ISSUE #198: Require JWT Bearer Token authentication
## Problem

Currently, the TES API specification does not mandate implementers to require any authentication scheme. The specification merely mentions the following [relevant paragraphs](https://github.com/ga4gh/task-execution-schemas/blob/117cd927aa119469ee80cedbd8940c968b81c5d6/openapi/task_execution_service.openapi.yaml#L52-L56) in its service description:

> Is is envisaged that most TES API instances will require users to authenticate to use the endpoints.
> However, the decision if authentication is required should be taken by TES API implementers.
>  
> If authentication is required, we recommend that TES implementations use an OAuth2 bearer token, although they can choose other mechanisms if appropriate.

Arguably, any production system offering access to compute resources MUST be protected from abuse by authentication (and authorization) policies, so mandating an authentication policy is prudent.

Moreover, enforcing at least one common authentication policy required to be supported by all implementations makes it considerably easier to design systems/environments in which multiple TES implementations are used concurrently, e.g., in federated TES networks.

## Overview

OpenAPI currently (version 3.x) supports the following [authentication (or authorization) schemes](https://swagger.io/docs/specification/authentication/):

- [**Basic**](https://swagger.io/docs/specification/authentication/basic-authentication/)
- [**Bearer**](https://swagger.io/docs/specification/authentication/bearer-authentication/)
- **Other HTTP schemes** as defined by [RFC 7235](https://tools.ietf.org/html/rfc7235) and [HTTP Authentication Scheme Registry](https://www.iana.org/assignments/http-authschemes/http-authschemes.xhtml)
- [**API keys**](https://swagger.io/docs/specification/authentication/api-keys/) in headers, query string or [cookies](https://swagger.io/docs/specification/authentication/cookie-authentication/)
- [**OAuth 2**](https://swagger.io/docs/specification/authentication/oauth2/)
- [**OpenID Connect Discovery**](https://swagger.io/docs/specification/authentication/openid-connect-discovery/)

Other GA4GH OpenAPI specifications define the following `securitySchemes`:

<html>
<div class="tg-wrap"><table>
<thead>
  <tr>
    <th>Specification</th>
    <th>SecuritySchemes</th>
    <th>Applied</th>
  </tr>
</thead>
<tbody>
  <tr>
    <td><a href="https://github.com/ga4gh-discovery/data-connect" target="_blank" rel="noopener noreferrer">Data Connect</a></td>
    <td><code>bearerAuth:</code><br><code>  type: http</code><br><code>  scheme: bearer</code></td>
    <td>Globally</td>
  </tr>
  <tr>
    <td><a href="https://github.com/ga4gh/data-repository-service-schemas" target="_blank" rel="noopener noreferrer">Data Repository Service (DRS)</a></td>
    <td><code>BasicAuth:</code><br><code>  type: http</code><br><code>  scheme: basic</code><br><code>  description: |</code><br><code>    A valid authorization token must be passed in the 'Authorization' header,</code><br><code>    e.g. "Basic ${token_string}"</code><br><code>BearerAuth:</code><br><code>  type: http</code><br><code>  scheme: bearer</code><br><code>  description:</code><br><code>    A valid authorization token must be passed in the 'Authorization' header,</code><br><code>    e.g. "Bearer ${token_string}"</code><br><code>PassportAuth:</code><br><code>  type: http</code><br><code>  scheme: bearer</code><br><code>  x-in: body</code><br><code>  bearerFormat: JWT</code><br><code>  description:</code><br><code>    A valid GA4GH Passport must be passed in the body of an HTTP POST request</code><br><code>    as a tokens[] array.</code></td>
    <td>Globally</td>
  </tr>
  <tr>
    <td><a href="https://github.com/ga4gh-discovery/ga4gh-service-info" target="_blank" rel="noopener noreferrer">Service Info</a></td>
    <td><code>bearerAuth:</code><br><code>  type: http</code><br><code>  scheme: bearer</code><br><code>  bearerFormat: JWT</code></td>
    <td>Globally</td>
  </tr>
  <tr>
    <td><a href="https://github.com/ga4gh-discovery/ga4gh-service-registry" target="_blank" rel="noopener noreferrer">Service Registry</a></td>
    <td><code>bearerAuth:</code><br><code>  type: http</code><br><code>  scheme: bearer</code><br><code>  bearerFormat: JWT</code></td>
    <td>Globally</td>
  </tr>
  <tr>
    <td><a href="https://github.com/ga4gh/tool-registry-service-schemas" target="_blank" rel="noopener noreferrer">Tool Registry Service (TRS)</a></td>
    <td><code>BEARER:</code><br><code>  type: apiKey</code><br><code>  name: Authorization</code><br><code>  in: header</code></td>
    <td>Individually for all endpoints</td>
  </tr>
  <tr>
    <td><a href="https://github.com/ga4gh/workflow-execution-service-schemas" target="_blank" rel="noopener noreferrer">Workflow Execution Service (WES)</a></td>
    <td>N/A<br><br><blockquote><em>"Users must supply credentials that establish their identity and authorization in order to use a WES endpoint. We recommend that WES implementations use an OAuth2 <a href="https://oauth.net/2/bearer-tokens/" target="_blank" rel="noopener noreferrer">bearer token</a>, although they can choose other mechanisms if appropriate. WES callers can use the <code>auth_instructions_url</code> from the <a href="https://ga4gh.github.io/workflow-execution-service-schemas/#/WorkflowExecutionService/GetServiceInfo" target="_blank" rel="noopener noreferrer"><code>service-info</code> endpoint</a> to learn how to obtain and use a bearer token for a particular implementation."</em></blockquote></td>
    <td>N/A</td>
  </tr>
</tbody>
</table></div>
</html>

> [Beacon](https://github.com/ga4gh-beacon/beacon-v2/) was omitted due to its complex structure of several OpenAPI specificatoin files, but does not appear to prescribe any authentication method

Bearer token authentication seems to be the consensus for those API specifcations that do prescribe authentication schemes, with the following exceptions:

- Next to bearer token authentication, [**DRS**](https://github.com/ga4gh/data-repository-service-schemas) requires compliant implementations to _also_ support basic and [GA4GH Passport](https://github.com/ga4gh-duri/ga4gh-duri.github.io/blob/82546d502c78ca341035a75c69f36edc8c50927a/researcher_ids/ga4gh_passport_v1.md)-based authentication schemes
- While the [**TRS**](https://github.com/ga4gh/tool-registry-service-schemas) `securityScheme` is _called_ `BEARER`, it actually mandates `apiKey`-based authentication; assuming that its name expresses its intention to be used equivalently to bearer token authentication, this is presumably a remnant from migrating the specification over from Swagger/OpenAPI 2.x, where [bearer authentication was not explicitly supported with its own scheme](https://swagger.io/docs/specification/2-0/authentication/)

## Proposed solution

Mandate implementers to support [JWT](https://datatracker.ietf.org/doc/html/rfc7519) [OAuth2 bearer tokens](https://www.rfc-editor.org/rfc/rfc6750) for authentication and apply it globally to all endpoints.

uniqueg:
Closes #198

The proposed change requires compliant TES implementations to implement JWT-based bearer token authentication.

This will make it easy for client applications to implement the popular [OpenID Connect](https://openid.net/developers/specs/) protocol to authenticate users and generate access tokens that could be used to give access to a varity of GA4GH API-backed microservices, following the [OAuth2](https://datatracker.ietf.org/doc/html/rfc6749) framework.

In fact, this is also the authentication/authorization flow suggested by the current GA4GH [Authentication and Authorization Infrastructure guidelines](https://github.com/ga4gh/data-security/tree/f5187cd3acc1f9d41bd724ca747e1bbc2f2e9f05/AAI).

As highlighted in the overview section of #198, bearer token authentication is also the current consensus across other GA4GH OpenAPI specifications that have defined at least one security scheme.

The suggestions to state `JWT` as the `bearerFormat` (which accepts arbitrary strings but mentions `JWT` in its [documentation](https://swagger.io/docs/specification/authentication/bearer-authentication/)) and to describe the expected behavior explicitly were included to strongly encourage implementers to follow JWT OAuth2 bearer token specifications expressly.

Of course, each implementation can still choose to support any number of alternative, or [additional](https://swagger.io/docs/specification/authentication/), security schemes.

@MattMcL4475

### DIFF
--- a/openapi/task_execution_service.openapi.yaml
+++ b/openapi/task_execution_service.openapi.yaml
@@ -49,14 +49,12 @@ info:
 
     ### Authentication and Authorization
 
-    Is is envisaged that most TES API instances will require users to authenticate to use the endpoints.
-    However, the decision if authentication is required should be taken by TES API implementers.
+    Implementers are required to implement the authentication schemes defined in
+    `components.securitySchemes`, but they are free to define and implement
+    additional schemes as per their requirements.
 
-
-    If authentication is required, we recommend that TES implementations use an OAuth2  bearer token, although they can choose other mechanisms if appropriate.
-
-
-    Checking that a user is authorized to submit TES requests is a responsibility of TES implementations.
+    Checking that a user is authorized to submit TES requests is the responsibility
+    of each implementation.
 
     ### CORS
 
--- a/openapi/task_execution_service.openapi.yaml
+++ b/openapi/task_execution_service.openapi.yaml
@@ -66,6 +66,8 @@ info:
 
 servers:
 - url: /ga4gh/tes/v1
+security:
+  - bearerAuth: []
 paths:
   /service-info:
     get:
@@ -246,6 +248,22 @@ paths:
               schema:
                 $ref: '#/components/schemas/tesCancelTaskResponse'
 components:
+  securitySchemes:
+    bearerAuth:
+      type: http
+      scheme: bearer
+      bearerFormat: JWT
+      description: |-
+        To access a resource to which this security scheme is applied, a valid JSON
+        JSON Web Token (RFC7519) MUST be passed as a bearer token (RFC6750) in the
+        'Authorization' request header, i.e., `Authorization: Bearer <TOKEN>`. If
+        the header is missing or malformed, or if the token is invalid, servers MUST
+        deny the request with a `401 Unauthorized` response.
+
+        Individual TES instances are encouraged to reference information on token
+        requirements and how a valid token can be obtained in the
+        `GET /service-info` response.
+
   parameters:
     view:
       name: view


denis-yuen (approved):
Thanks for the heads-up!

openapi/task_execution_service.openapi.yaml
```diff
@@ -246,6 +246,22 @@ paths:
               schema:
                 $ref: '#/components/schemas/tesCancelTaskResponse'
 components:
+  securitySchemes:
+    bearerAuth:
+      type: http
+      scheme: bearer
+      bearerFormat: JWT
```
patmagee, line 253: Requiring JWT for an access token may be a bit overly prescriptive. The OIDC Standard does not require access tokens to be in the JWT format, only ID tokens so systems are free to implement access tokens as Opaque Strings.

Less important then the format of the token is the flow that is involved. Is there a way to define OIDC as the flow that is required?

MattMcL4475: @patmagee do you think both BearerAuth and OpenID should be supported, or just OpenID?

https://swagger.io/docs/specification/authentication/

```
components:
  securitySchemes:
    OpenID:
      type: openIdConnect
      openIdConnectUrl: https://example.com/.well-known/openid-configuration
```

MattMcL4475: fyi @uniqueg @patmagee :

_The GA4GH TES (Global Alliance for Genomics and Health Task Execution Schema) is designed for interoperability and often used in multi-tenant, distributed, and security-sensitive environments. Here are some considerations for each security scheme:

OpenID:
Since you want to support OpenID, including this is a must. OpenID Connect is well-suited for scenarios where identity information and distributed single sign-on are required.

yaml
Copy code
OpenID:
  type: openIdConnect
  openIdConnectUrl: https://example.com/.well-known/openid-configuration
BasicAuth:
Basic Authentication is simple but less secure. In a setting where advanced features like single sign-on or multi-factor authentication are important, BasicAuth falls short.

BearerAuth:
Bearer Tokens are a common way to handle API authentication. They can work well with OpenID Connect if the tokens are JWTs issued by an OIDC Identity Provider. You might include this for more simplistic clients that aren't OIDC-compliant but can handle JWTs.

ApiKeyAuth:
API keys are straightforward for client-to-server authentication but aren't designed for scenarios where you need to authenticate end-users or deal with federated identity providers.

OAuth2:
OAuth2 is more flexible than OIDC and can provide more granular permissions via scopes. OIDC is actually built on top of OAuth2, so if you're implementing OIDC, you're implicitly using OAuth2. Including a separate OAuth2 security scheme could be redundant, but it might be useful if you have different services or levels of access requiring OAuth2's more extensive capabilities.

Recommendations:
Mandatory: Since you wish to support OpenID, the OpenID entry should definitely be there.

Conditional: BearerAuth can stay if you envision clients that will use simple Bearer Tokens instead of going through the full OIDC flow.

Optional: OAuth2 can be included for use-cases that need more granular permissions not covered by your OIDC scopes or for clients that are specifically OAuth2-compliant but not OIDC-compliant.

Discouraged: BasicAuth and ApiKeyAuth are generally not recommended for sensitive, distributed, or multi-user environments.

Based on your specific needs, you can decide to keep some, all, or none of the other schemes in addition to OpenID._

uniqueg: Good point @patmagee. Indeed, TES and similar microservices would receive access tokens, not ID tokens, and (OAuth2-compliant) access tokens _could_ indeed be opaque as well.

However, the main point for this issue and PR is to form a basis on which different TES implementations would be interoperable in terms of authN and authZ (we can worry about that later, it's much more complex), so as to be able to realize multicloud use cases. And if every implementation uses a different token format, you would still always need agreements between client and server, which is not scalable (at least that is my understanding). Therefore I feel that if we do not explicitly request implementers to support JWT-based bearer tokens, we wouldn't have won much.

Also, I think the issue is only relevant for _public_-facing APIs. If organizations choose to use, e.g., TES internally, there is no strong need to be fully compliant. The same is true for any organization that deliberately chooses not to support one requirement or another for their public-facing APIs, for whatever reason. Nobody is required to implement the entirety of a specification - it would just limit the interoperability. In that way I feel that putting this in the specs prescriptively is for implementers caring about a wide level of interoperability, in which case JWT bearer tokens make sense (I feel).

As for the flow: OpenAPI supports [OAuth2](https://swagger.io/docs/specification/authentication/oauth2/) where you can then specify flows. However, I think this is going too deep into AuthZ for the time being. As far as I can tell from previous discussions on the topic (work order tokens, Google's macaroons, the complex use cases and requirements summarized [here](https://docs.google.com/document/d/1wiCLPFFDw7gczSLYC7LRtvPHI4GnfRjv/edit#heading=h.gjdgxs)), we are not at all sure what an "interoperable access control" would look like, and whether it would or _could_ be compliant with OAuth2 (e.g., it would appear to me that adding Passports/Visas to the payload of a `POST` endpoint is _not_ - apart from being a rather serious REST violation). So while I would actually _love_ to be even more prescriptive here, I'm not sure that that wouldn't come back to haunt us (I feel moving from a bearer to an OAuth2 scheme is easier than the other way around).

As for OIDC, OpenAPI supports the [OIDC Discovery](https://swagger.io/docs/specification/authentication/openid-connect-discovery/) scheme. But as I said above, I think TES and similar services would need to process access tokens, not ID tokens, so I don't think it makes sense to use that scheme here or elsewhere among the Cloud APIs (although through the strong encouragement to support JWT bearer tokens, it is pretty much implied that access tokens would be generated through OIDC on the client side, and not, e.g., SAML).

All in all, I feel that the suggested PR, including the explicit mention of the JWT format (which aligns well with Passport and the GA4GH AAI policies), is a good middleground to prepare for a integrated interoperable AuthN/AuthZ policy across GA4GH API-powered mulitclouds, without necessarily committing to OAuth2 or any other standard except JWT.

patmagee: @uniqueg I can see what you are trying to achieve here.

While jwts do simplify server code (ie you just need to understand how to parse them and all the information is there), it does not simplify trust. 

In a real world setting a single JWT is likely only usable for a single TES api, since the tes service would need to trust the upstream issuer, and the token should ideally be scoped to a specific audience.

If that is that case the calling service will need to get a token for each of the downstream tes services anyway. If the flow is discoverable via a discovery endpoint then this is possible and the specific payload of the token is really not important.

If the intention is that a single token will work on all downstream tes servers, I can see there being issues there unless there is a priori trust with the issuer. At which point the specific format also does not matter since all services in the system trust the same issuer

uniqueg: Yes, the use case is for using the same access token across a network of multiple TES instances and/or other GA4GH (Cloud) API-powered microservices.

But how do I know who the issuer is if the token is opaque? It may have been passed along to TES via a WES, for example. And how do I get information on whose behalf the resource is requested, for access control and accounting?

As far as I can see, JWTs would be able to address that, but not opaque tokens.

Imagine that my cloud has a service registry of WES and TES instances it trusts, then users can grant permission individually to narrow down that list to the instances _they_ trust (GDPR). Based on that an access token could be furnished with the required audience. Similarly, the TES instances would add the issuer of the tokens for my cloud to their list of trusted authorities.

Both adding a TES instance (or entire network of them) to the service registry, and adding an issuer to the list of trusted authorities with a given TES instance (or a network thereof) could be simplified to the extent that it could be a simple review process (with periodic re-review if necessary).

Or am I missing something here?

Of course this is just for authN. Even with this PR, organizations or individual instances could still could make it as easy or hard for users to consume resources as they would like.  How to simplify users actually getting access to individual resources (e.g., via a Visa-like mechanism) is a different story.

uniqueg: Btw, this is not just hypothetical. We are actually building exactly such a system, with service registry, WES and TES gateways for workflow and task distribution, respectively, and individual WES and TES instances of different implement that are run by individual trusted organizations across Europe/ELIXIR. And we are running exactly into the problem that different implementations/instances are already requiring different authentication schemes, requiring us to integrate compatibility layers into our gateways which need to process secrets, like basic auth credentials for individual services.

It's all possible, but not very scalable, especially if in the future we want to share compute not just across ELIXIR Nodes, but also with, say, NIH, DDBJ etc.

patmagee: >  Yes, the use case is for using the same access token across a network of multiple TES instances and/or other GA4GH (Cloud) API-powered microservices.

I guess for us this would never happen. Reusing an access token across services within a network especially across organizational boundaries is generally considered a bad security practice.

You end up with an excessively powerful token that has a fairly large attack surface. 

The ideal scenario is for each service to accept a token with a specific and unique `audience` which means that token is only usable on that service. 

For us, we accomplish this using token exchange to take the users Id token or access token and exchange it for a token that can work on the target service with the designated issuer. Hence the actual contents of that token do not really matter to us.

patmagee: Using token exchange each tes can have its own issuer it trusts, so long as the gateway WES/TES knows about it and has a client with that issuer that can perform a token exchange.

This has the benefit of using the actual user permissions

uniqueg: This sounds like an interesting pattern indeed. Would you care to elaborate on who calls whom for the token exchange, and how you think a TES could broadcast its trusted issuer so that the gateway (or WES, or whatever upstream client that may be interested in talking to multiple TES instances) could negotiate the exchange for any given issuer once the preferred TES has been determined? Or a reference with a good explanation and ideally a call diagram specifically for the cross-organizational use case? I've been looking for more information on this exact problem for a long time, but failed to make any real headway.

uniqueg: Also would like to tag @martin-kuba here

patmagee: @uniqueg for a commercial example you can look at Google Clouds [Workload Identity Federation](https://cloud.google.com/iam/docs/workload-identity-federation) for inspiration.

In terms of TES, I think we really have everything we need to make this possible

Please note there are definitely multiple ways to achieve this, but this is just one:


1. The service info would need need to expose authorization information such as the token uri, or better yet the well known configuration endpoint
2. The TES/WES gateway would need to know about the existence of the the tes instance and it's auth info (retrieved from the well known config)
3. In a perfect world, each issuer will also expose an OAuth endpoint for client registration
4. In reality, the gateway would likely need to have a client with the target tes pre-configured with the appropriate grant types configured (in this case the token exchange grant type)
  - the issuer may need to be configured (as is the case with google) to "know" how to process the up stream token
5. When a request comes into the gateway, the gateway takes the submitted access token and uses the appropriate client to exchange the token for a new one that is useable with the target tes instance


If this seems complex, it's because it kind of is. We are dealing with federated trust here and that is great, but it also introduces complexity.

Brining it back to your original PR (this is a bit orthogonal to it), I think this probably illustrates why requiring JWT may be a bit restrictive as opposed to just requiring bearer auth.

PR #200, opened 2023-10-24
Feature: add TesTask.state_description

MattMcL4475:
Currently, when Azure Batch has no quota available, a TES Task in TES on Azure will stay in the `INITIALIZING` state indefinitely until quota becomes available.  TES needs a way to inform the caller why this is the case, so that the caller can update the UI with this additional information, and the user or IT admin knows they need to submit an Azure Support Request to increase their quota.  Otherwise, they don't have visibility into why the task is not progressing.

### DIFF
--- a/openapi/task_execution_service.openapi.yaml
+++ b/openapi/task_execution_service.openapi.yaml
@@ -743,8 +743,7 @@ components:
         state_description:
           type: string
           description: |-
-            OPTIONAL. Additional information about the state of the task.
-            This may be used to provide more information about the state of the task,
+            OPTIONAL. Additional information about the state of the task,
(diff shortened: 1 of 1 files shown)


MattMcL4475:
@patmagee what are your thoughts on the best way to handle this?

patmagee:
@MattMcL4475 i would be in favour of a fail fast model. I think waiting for quota for a given period of time is okay, but beyond a "reasonable limit" the tes task should fail.

I propose a new state `INSUFFICIENT_RESOURCES` to be treated as a failure state. If returned it would be highly informative to the end user why something failed. Combine that with a message for the failure and that should allow diagnosis of most failures

patmagee:
I think this state would fit well onto a WES workflow as well btw

PR #219, opened 2024-10-08
Update main to spec standards for 1.1

vsmalladi:
* Update master branch to lastest stable spec TES 1.1.
* Also update build of documentation to relay on the main site.


uniqueg (commented):
The doc build still seems broken? Is that expected?

Also, a small PR description might help in understanding better what's being done here and with what purpose. Is this for the docs build?

Also, I don't quite get the title - why the "for gh-actions"? Aren't we merging develop into master whole here, including spec changes?

vsmalladi:
@uniqueg should be better now

.github/workflows/build.yml
```diff
@@ -0,0 +1,44 @@
+name: Build OpenAPI docs
+on:
+  - push
+  # - pull_request
```
uniqueg, line 4: I'm not sure. Isn't what we want more like:

```yaml
on:
  pull_request:
    types: [opened, synchronize, reopened]
  push:
    branches:
      - master
```

?

This would:
- Run exactly one instance of the workflow - with only the `docs-build` job - whenever a pull request is opened (against any target branch), updated (regardless of the number of commits in the update) or reopened after being closed.
- Run exactly once for a push against `master`, i.e., after a PR against `master` has been merged; in this case, due to `if: github.ref == 'refs/heads/master', also the `deploy` job would be run.

I think this would be less broad/more specific.

Alternatively, perhaps remove the `# - pull_request` line.

.github/workflows/build.yml
```diff
@@ -0,0 +1,44 @@
...
+      - run: npm install -g gh-openapi-docs
+      - name: Check out repository code
+        uses: actions/checkout@v4
+      - run: gh-openapi-docs
+      - name: Deploy 🚀
+        uses: JamesIves/github-pages-deploy-action@v4
+        with:
+          branch: gh-pages
+          folder: .
+  deploy:
```
uniqueg, line 24: Could we somehow avoid the redundancy with `docs-build`? Seems like only two parameters are added (`target-folder` and `clean`). Perhaps we could check the branch and operation we are on and make sure that we set these accordingly?

Alternatively, we could maybe negate the `if: github.ref == 'refs/heads/master'` in `docs-build`, so that we aren't running two slightly different deployment jobs one after another after merging into `master`?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,47 @@
+name: Lint and validate OpenAPI specs
+on:
+  - push
+  - pull_request_target
```
uniqueg, line 4: See above for a suggestion on a potentially better definition of workflow trigger events.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,47 @@
...
+  lint:
+    name: Lint OpenAPI definition
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out head branch
+        uses: actions/checkout@v4
+      - name: Run OpenAPI Lint Action
+        uses: nwestfall/openapi-action@v1.0.2
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
```
uniqueg, line 15: Isn't this risky, together with `pull_request_target` above? If I understand this correctly, people creating a PR from a fork might be able to read the `GITHUB_TOKEN` and could mess with the repo as per the permissions the token has.

On the other hand, with the current configuration, this job wouldn't run if triggered from a fork (which is our usual workflow). Perhaps we should check whether we can lint the workflow without the secret and then use `pull_request` instead of `pull_request_target`? It seems to me that a linting check shouldn't need to require specific permissions (just like the `diff` and `validate` jobs don't need it).

README.md
```diff
@@ -1,9 +1,12 @@
-<img src="https://www.ga4gh.org/wp-content/themes/ga4gh-theme/gfx/GA-logo-horizontal-tag-RGB.svg" alt="GA4GH Logo" style="width: 400px;"/>
+<img src="https://www.ga4gh.org/wp-content/themes/ga4gh/dist/assets/svg/logos/logo-full-color.svg" alt="GA4GH Logo" style="width: 400px;"/>
+
+![release_badge](https://img.shields.io/github/v/tag/ga4gh/task-execution-schemas)
+
 
 Task Execution Service (TES) API
 ======================================
-<sup>`master` branch status: </sup>[![Build Status](https://travis-ci.org/ga4gh/task-execution-schemas.svg?branch=master)](https://travis-ci.org/ga4gh/task-execution-schemas?branch=master)
-<a href="https://raw.githubusercontent.com/ga4gh/task-execution-schemas/master/openapi/task_execution_service.openapi.yaml"><img src="http://online.swagger.io/validator?url=https://raw.githubusercontent.com/ga4gh/task-execution-schemas/master/openapi/task_execution_service.openapi.yaml" alt="Open API Validator" height="20em" width="72em"></A>
```
uniqueg, line 6: Kind of a pity to miss out on them, no?

I think for the build status, we could update the badge to use GitHub Actions instead of Travis.
For the OpenAPI validation, it still seems to work, but maybe we should replace `http` with `https`.

Also, instead of the HTML, we should just use markdown syntax - and move them above the header, together with the release badge.

While we are at it, we could also add a license badge.

README.md
```diff
@@ -1,9 +1,12 @@
-<img src="https://www.ga4gh.org/wp-content/themes/ga4gh-theme/gfx/GA-logo-horizontal-tag-RGB.svg" alt="GA4GH Logo" style="width: 400px;"/>
+<img src="https://www.ga4gh.org/wp-content/themes/ga4gh/dist/assets/svg/logos/logo-full-color.svg" alt="GA4GH Logo" style="width: 400px;"/>
+
+![release_badge](https://img.shields.io/github/v/tag/ga4gh/task-execution-schemas)
```
uniqueg, line 3: I think it would be nice to add a URL target for the badge. In this case, I think it could bring us to the releases page: https://github.com/ga4gh/task-execution-schemas/releases

README.md
```diff
@@ -1,9 +1,12 @@
...
+
 
 Task Execution Service (TES) API
 ======================================
-<sup>`master` branch status: </sup>[![Build Status](https://travis-ci.org/ga4gh/task-execution-schemas.svg?branch=master)](https://travis-ci.org/ga4gh/task-execution-schemas?branch=master)
-<a href="https://raw.githubusercontent.com/ga4gh/task-execution-schemas/master/openapi/task_execution_service.openapi.yaml"><img src="http://online.swagger.io/validator?url=https://raw.githubusercontent.com/ga4gh/task-execution-schemas/master/openapi/task_execution_service.openapi.yaml" alt="Open API Validator" height="20em" width="72em"></A>
+
+This repository is the home for the schema for the GA4GH Task Execution API. The Goal of the API is to provide a standardized way to executing batch execution tasks.
```
uniqueg, line 9: - We are using GA4GH before spelling it out first
- It should be Task Execution Service API, not Task Execution API
- The description of the Cloud Work Stream later on comes out of nowhere, it is not clear that the TES API is defined as part of that work stream
- Links for GA4GH and Cloud Work Stream are outdated

How about this?

> This repository is the home for the schema of the Task Execution Service (TES) API defined by the [Cloud Work Stream](https://www.ga4gh.org/work_stream/cloud/) of the [Global Alliance for Genomics and Health (GA4GH)](https://ga4gh.org/). The goal of the API is to provide a standardized way to executing batch execution tasks.

README.md
```diff
@@ -27,9 +30,17 @@ API Definition
 --------------
 
 See the human-readable [Reference Documentation](https://ga4gh.github.io/task-execution-schemas/docs/)
```
uniqueg, line 32: This statement misses a full stop at the end.

README.md
```diff
@@ -27,9 +30,17 @@ API Definition
...
-and the [OpenAPI YAML description](openapi/task_execution_service.openapi.yaml). You can also explore the specification in the [Swagger Editor](https://editor.swagger.io/?url=https://ga4gh.github.io/task-execution-schemas/openapi.yaml).
+You can also explore the specification in
+the [Swagger Editor](https://editor.swagger.io/?url=https://ga4gh.github.io/task-execution-schemas/openapi.yaml)**
+*Manually load the JSON if working from a non-develop branch version.* Preview documentation from
```
uniqueg, line 35: I think there is some markdown mixup here. See: https://github.com/ga4gh/task-execution-schemas/blob/develop/README.md

README.md
```diff
@@ -27,9 +30,17 @@ API Definition
...
+the [gh-openapi-docs](https://github.com/ga4gh/gh-openapi-docs) for the development
+branch [here](https://ga4gh.github.io/task-execution-schemas/preview/develop/docs/index.html)
+
+> All documentation and pages hosted at 'ga4gh.github.io/task-execution-schemas' reflect the latest API release from
+> the `master` branch. To monitor the latest development work, add 'preview/\<branch\>' to the URLs above (e.g., '
+> ga4gh.github.io/task-execution-schemas/preview/\<branch\>/docs'). To view the latest *stable*
+> development API specification, refer to the `develop` branch.
```
uniqueg, line 42: Are we sure this is still accurate with the new build pipeline?

README.md
```diff
@@ -27,9 +30,17 @@ API Definition
 --------------
 
 See the human-readable [Reference Documentation](https://ga4gh.github.io/task-execution-schemas/docs/)
-and the [OpenAPI YAML description](openapi/task_execution_service.openapi.yaml). You can also explore the specification in the [Swagger Editor](https://editor.swagger.io/?url=https://ga4gh.github.io/task-execution-schemas/openapi.yaml).
+You can also explore the specification in
+the [Swagger Editor](https://editor.swagger.io/?url=https://ga4gh.github.io/task-execution-schemas/openapi.yaml)**
+*Manually load the JSON if working from a non-develop branch version.* Preview documentation from
+the [gh-openapi-docs](https://github.com/ga4gh/gh-openapi-docs) for the development
+branch [here](https://ga4gh.github.io/task-execution-schemas/preview/develop/docs/index.html)
```
uniqueg, line 37: Also misses full stop. And perhaps we could make a full sentence out of this.

README.md
uniqueg, line 42: I can't comment on this, but I think the section "TES Compliant Implementations" has some issues:
- I don't think it makes sense to use note/quote markdown for an entire section. Rather, I would emphasize (in bold) the most important bit or bits.
- The wording is strange and there is a comma missing.

How about:

> Complying with GA4GH policies and regulations, security reviews are performed for each major version release of the API. However, **no security guarantees are given for any implementation of the API, including any linked to from this page or the accompanying documentation**. Use such implementations at your own risk and make sure to arrange for a security audit of your application to ensure compliance with any applicable regulatory and security requirements, especially where personal data may be at issue.

README.md
```diff
@@ -39,12 +50,31 @@ TES Compliant Implementations
...
+> Compatibility is assumed based on available documentation and limited tests performed on latest versions of implementations available as of December 2020.
+
+|               | cwl-tes       | Cromwell  | Nextflow |
+| -----------   | ------------- | --------- | ---------|
+| **Funnel**    | [Compatible](https://github.com/ohsu-comp-bio/cwl-tes) | Compatible | [Compatible](https://www.nextflow.io/docs/latest/executor.html#ga4gh-tes) |
+| **TESK**      | [Compatible](https://github.com/ohsu-comp-bio/cwl-tes/pull/25) | [Compatible](https://cromwell.readthedocs.io/en/stable/backends/TES/) | [Compatible](https://github.com/EMBL-EBI-TSI/tesk-core/pull/19) |
+| **TES Azure** | Not tested | [Compatible](https://github.com/microsoft/CromwellOnAzure) | [Compatible](https://www.nextflow.io/docs/stable/executor.html#ga4gh-tes) |
 
 
 TES Service Examples
```
uniqueg, line 80: I think the parts below could use some work:

TES Service Examples section:
- The `name` field in the "Create a task" section includes unncessary whitespace
- Invalid JSON looks bad (highlighted in dark red)
- The `POST` example is incomplete and it isn't clear that the following JSON is a response (rather than, say, a payload)
- I've seen some more typos and smaller issues, probably best to address in a separate issue/PR

Possible Future Enhancements section:
- I would remove this entirely, or rephrase this and add a link to the issue board instad.

Finally, I think it would be nice to include sth like a "Governance" section - which makes clear a bit how we work. People will probably want to know how their proposed changes can make it into the specs (including timelines).

openapi/task_execution_service.openapi.yaml
uniqueg, line 1: Not reviewing this, as this is what the version is.

uniqueg (changes requested):
Looks good @vsmalladi - but I think we should use the opportunity to iron out a few issues, including legacy issues.

What do you think?

I'll try to help with this over the next weeks.

.github/workflows/build.yml
vsmalladi: I think we want the `pull_request` line since then we can have documentation for `master`, `development` and other branches to see issues in the documentation. Thoughts?

.github/workflows/build.yml
vsmalladi: Ya i think this can be simplified. Let me see if i can fix this.

README.md
vsmalladi: Good idea. Updating.

README.md
vsmalladi: Good idea. Working on it.

README.md
vsmalladi: Great idea

README.md
vsmalladi: Yep that is true. Double checked it https://ga4gh.github.io/task-execution-schemas/preview/develop/docs/

vsmalladi: Good idea. Better workind

uniqueg: Thanks. So we can close this convo, right? If so, please go ahead.

.github/workflows/build.yml
uniqueg: I agree. But in the current workflow it's commented out in favor of every `push` (which is very broad) - in which case we really don't need `pull_request` at all.

My suggestion would run the workflow for every PR (opened, updated or reopened), but not for branches for which no PR exists. PLUS it would run _after_ merging changes into `master`. Would this be sufficient in your opinion, @vsmalladi?

vsmalladi: Ya i agree. Let me change that

.github/workflows/ci.yml
vsmalladi: Testing a new plugin for validation `mhiew/redoc-lint-github-action@v4`

.github/workflows/build.yml
vsmalladi: Fixed

.github/workflows/ci.yml
vsmalladi: Fixed

uniqueg (approved):
LGTM

PR #221, opened 2024-10-21
Update logo to render in html documentation

vsmalladi:


uniqueg (approved):
Bit of a pity that the w3id link that was made to remain stable is dead 😅

PR #224, opened 2024-10-29
Update build.yml

uniqueg:
The change ensures that workflow runs are only triggered when commits are pushed to the protected branches `main`, `develop` and those starting with `release` - where changes are expected to be merged through PRs. For building docs for feature branches, the `pull_request` trigger will be sufficient.

This setup avoids repeat runs through both the `push` and `pull_request` triggers in most scenarios.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -3,6 +3,7 @@ on:
   push:
     branches:
       - main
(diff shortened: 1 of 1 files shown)


vsmalladi (approved):
LGTM

PR #227, opened 2024-10-29
PR comment for OpenAPI Diff output

uniqueg:
- Do not fail when OpenAPI Diff action errors
- Instead, create PR comment with error log
- Currently does not work because of permission issues (403 on trying to create comment); may need to set PAT in github-script action, but don't have permissions to do that for this repo
- Otherwise works fine

Produces output like this:

[diff.md](https://github.com/user-attachments/files/17564426/diff.md)


vsmalladi (approved):

vsmalladi:
@uniqueg  what do we want to do with this PR?

vsmalladi:
@uniqueg still issue with permissions. But got a workaround to upload the artifact with error log.
[diff-artifacts (2).zip](https://github.com/user-attachments/files/18542064/diff-artifacts.2.zip)

vsmalladi:
@uniqueg  I think i fixed this.

vsmalladi (approved):
All changes look good. Merging in now.

PR #244, opened 2025-12-02
Implement GetCapabilities endpoint in OpenAPI spec

### LINKED ISSUE #188: How do we define the technical capabilities of a given TES API
TES provides an abstraction layer on-top of the smallest unit of work in the workflow execution stack and interfaces directly with some sort of compute infrastructure. At the moment, user's of the TES API need to know and understand the technical constraints of the underlying compute environment in order to know what sort of capabilities a given TES api has. This is a problem in so far as it leaks the implementation details and requires the user to know information prior to using the API.

IF we push the boundaries of WES and TES the natural conclusion is federating work submitted to WES across different TES backends that fit the mould of the requested task resources/capabilities. This implies A LOT of machine -> machine interaction where it is a WES api determining where to send work. In order to accomplish this we really need a way to describe the complete list of technical capabilities of a TES backend. #186 is a good example of a specific technical ability that would need to be described in some way. Other examples would be GPU support (and what line GPUS), the range of CPU's or supported data types etc etc.

You could imagine that a workflow run through WES could take each individual task (in CWL/WDL perlance) and use heuristics to map it onto a particular TES backend. I know that some work has been done by @uniqueg and his team on building a Gateway TES that does a similar role, but I wonder what would be required to make any TES api able to participate in this gateway approach

### LINKED ISSUE #206: feature request: capabilities
I suggest that we add an endpoint that informs clients of the capabilities of the execution service (similar to [capabilities](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#capabilities)) in the LSP specification. For example, a client should be able to, at minimum, figure out the following:

- What is the total capacity a service can provide (including pending jobs, even if it is unbounded).
- What is the expected concurrent capacity that a service can provide (just running jobs, even if it is unbounded).
- Does the service support preemptible instances?

vsmalladi:
Added capabilities endpoint to retrieve TES backend features. Closes #206 
Closes #188

### DIFF
--- a/openapi/task_execution_service.openapi.yaml
+++ b/openapi/task_execution_service.openapi.yaml
@@ -266,6 +266,38 @@ paths:
       security:
         - BearerAuth: []  # Endpoint secured with Bearer token
         - BasicAuth: []    # Alternatively, secured with Basic Auth
+  /capabilities:
+    get:
+      summary:GetCapabilities
+      description: |
+        Returns information about compute, hardware, accelerator, and
+        machine-topology features supported by this TES backend.
+
+        All fields are OPTIONAL. Implementations MAY return only partial
+        or high-level information.
+      operationId: GetCapabilities
+  responses:
+        "200":
+          description: TES capability definition
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/tesCapabilities"
+              example:
+                cpu_cores: 32
+                ram_gb: 256
+                disk_gb: 2000
+                num_machines: 8
+                accelerators:
+                  - type: GPU
+                    count: 4
+                  - type: TPU
+                    count: 2
+                chip_type: x86_64
+                fpga: true
+      security:
+        - BearerAuth: []  # Endpoint secured with Bearer token
+        - BasicAuth: []    # Alternatively, secured with Basic Auth
 components:
   securitySchemes:
     BearerAuth:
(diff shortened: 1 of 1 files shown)


vsmalladi:
Response would be like this

```
{
  "cpu_cores": 32,
  "ram_gb": 256,
  "disk_gb": 2000,
  "num_machines": 8,
  "accelerators": [
    {
      "type": "GPU",
      "count": 4
    },
    {
      "type": "TPU",
      "count": 2
    }
  ],
  "chip_type": "x86_64",
  "fpga": true
}
```

vsmalladi:
Add docker or other container resources.

andrew-nimbus:
I also wonder if we want to focus an initial version of the capability on providing GPU metadata (in addition to some baseline CPU/disk/mem inventory as well)--one would probably want things like the vendor, model, GPU mem per device, driver type and version, GPU execution model (e.g., exclusive vs. partitioned), runtime version, and the types of containers supported (which you note in the comment--indicating singularity support would likely be esp useful)--right? I'm trying to think about how irritating it can be to match a container to an accelerator environment.

Along these lines, do we want to expand the GPU metadata (to potentially target an eventual PoC implementation) and drop the FPGA flag for the moment (as it's unclear to me how to use this boolean in practice...but ive never used FPGA accelerators before).

andrew-nimbus:
On another note, we'll want to make clear in the updated documentation that the initial goal here is to facilitate helping jobs to run on the hardware (esp. accelerators)--scheduling and optimization are for later.

I.e., this is a static _inventory_ (i.e., "is there a GPU accessible to the TES endpoint") but not focused on _status_ (e.g., "is the GPU available now?"). Is the implicit assumption that the TES server is effectively managing current state via its set of workers at the moment? I think this is perfectly reasonable, but it would be worth being explicit about it in any updated documentation.

uniqueg (commented):
I'm all for this, but would prefer for this to be hosted at `/service-info` (as a product-specific extension of the [Service Info API](https://www.ga4gh.org/product/service-info/)), so that it becomes discoverable through the [GA4GH Service Registry API](https://www.ga4gh.org/product/service-registry/).

Also, perhaps we should think of a more generic/extensible format to encourage plugin development by the community.

Finally, I think we should consider coordinating this with other GA4GH products within the Cloud WS, or across the Cloud and Discovery WS (which together govern all, or almost all, of GA4GH's OpenAPI products).

A possible approach and example use cases are outlined in https://github.com/ga4gh/TASC/issues/45.

PR #204, opened 2024-07-12
fix: zones example in data schema

JaeAeich:
fixes - #203

PR #246, opened 2026-02-12
fix: Make resources.zones example an array instead of string

BKDaugherty:
The documentation says this should be an array, but the example just has a single string.

Screenshot 2026-02-11 at 5 10 04 PM
Screenshot 2026-02-11 at 5 10 12 PM

[Here's the gh-pages from my fork](https://bkdaugherty.github.io/task-execution-schemas/preview/develop/docs/#tag/TaskService/operation/CreateTask), which has it as an array!

### DIFF
--- a/openapi/task_execution_service.openapi.yaml
+++ b/openapi/task_execution_service.openapi.yaml
@@ -641,7 +641,8 @@ components:
             priorty queue to which the job is assigned.
           items:
             type: string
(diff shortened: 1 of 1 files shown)


BKDaugherty:
Duplicate of https://github.com/ga4gh/task-execution-schemas/pull/204.

